PR #99, opened 2026-04-08
Ingest new NGED data from S3

### LINKED ISSUE #59: Select only the 33 timeseries in the trial area
~~But continue downloading live NGED flows for all substations~~

### LINKED ISSUE #68: Ingest NGED's new S3 data
NGED have substantially changed how they deliver their power data to us. We had previously been getting data as CSV files from CKAN. Now the data will be available as JSON files in an S3 bucket.

The new data on S3 has a very different file format. All the data we need for a given substation is included in each JSON file (e.g. the geographical location, and the substation name, and the power data). The metadata will be repeated when the new JSON files are published by NGED every 6 hours.

Let's keep our local data structure the same as when we were reading from CKAN: i.e. we'll have a separate Parquet for the substation metadata, and we'll store the power data in a Delta table. All our local will still be strictly controlled by our data contracts.

The new JSON data includes the start and end time of each power reading. Let's simplify this by only storing the end time of each recording, and making sure that we document that our data is period-ending.

Note that this new data is not just about _primary substations_. This new data is _mostly_ about primary substations. But also includes data from bulk supply points and grid supply points, customer data (e.g. large metered PV sites). NGED now use a `time series ID` to uniquely identify each time series. For example, one time series might be the power flow for a specific primary substation. Another time series might be the power flow for a customer's PV site.

There are separate archival files (which contain several years of data) and live files (which contain only something like a rolling 2 weeks of data). The live files will be updated every 6 hours.

We have to create forecasts for all these different types of assets.

The workflow should be something like this:
1. When we first download the multi-year archive JSONs, we populate our metadata
2. Then, when we download the updated data every 6 hours, we check the new metadata against our old metadata. If an update is necessary then go ahead and perform the update, but also notify the user somehow.

There's lots to do! The changes we need to make include:
 
- Explore the new data in S3. Understand its structure. Come up with a plan for how to implement downloading, and to implement a data contract to validate the incoming JSON data. The downloading and processing data should all live in  Implement code in `packages/nged_data`, including the Dagster assets.
- Remove all the old code for downloading data from CKAN. Remove any mentions CKAN in the docs or the settings.
- Implement checks for the data. The archival power data starts as all zeros. We need to remove that data before appending it to our local Delta table.

Some questions:

- How should we partition our Dagster asset to download NGED power data given that NGED's S3 contains archival data files, and files that are updated every 6 hours. So our currently daily partitions probably don't make sense.

### LINKED ISSUE #100: Remove deprecated NGED CKAN code
Once we're confident that NGED's S3 data pipeline is working well, then delete all the CKAN code, including:

- The entire `nged_data` package.
- All the old CKAN Dagster assets.
- Anything in the data contracts that's only used in the CKAN data. e.g. different substation names from the location data versus the live substation data.

## TODO
- [x] Implement code to ingest NGED's new JSON data
- [x] Remove code that ingested CKAN data
- [x] Why does PowerTimeSeries schema have `startTime`? Can't we just store `endTime`, and ensure the incoming data is half-hourly?
- [x] Rename `value` to `power` in the XGBoostFeatures, and elsewhere?
- [x] Maybe rename `end_time` to `valid_time` in `PowerTimeSeries` to match the use of `valid_time` in the NWP data? Or maybe it's nice to be explicit that `end_time` is the end time of the half-hour period???
- [x] Go through the sample archive JSON data, and find the allowed values (e.g. for substation_type), and tighten up the data contracts.
- [x] Rename `packages/nged_json_data` to `nged_timeseries_data`? (assuming we use a different package for the non-time-series data, like the adjacency matrix?) Or maybe to `nged_data`?
- [x] Run the pipeline on the JSON data from sharepoint. Save Delta table & Parquet to disk. Visualise in dashboard.
- [x] Rename `flows_30m` to `power_time_series`
- [x] Check the pipeline for creating the `h3_res_5` column in the `time_series_metadata.parquet`, and - crucially - make sure it's not over-complicated. I think there might be some silly code in the model training that fills in the H3 if it's missing. But it shouldn't be missing!
- [x] Run full pipeline (including model training & eval) on NGED's JSON archive
- [x] Why are we getting spurious periods of zeros in the cleaned power data?
- [x] #105
- [ ] Pull data from NGED's S3 bucket.
- [ ] Final automated code review
- [ ] Manual review code 

## Related
- #99
- #68

JackKelly:


JackKelly:
This still needs a thorough human review. But I'll do those reviews in follow-on PRs

PR #183, opened 2026-06-24
ML dagster phase 1

JackKelly:
Implements Phase 1 of #171. Follows on from #182.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -81,8 +81,21 @@ Identity is split across two levels. **Model-family identity** — `MODEL_NAME` 
 - **Polars only** — pandas is strictly forbidden. Use `pl.LazyFrame` and only `.collect()` when necessary.
 - **Patito** for all DataFrame schema definitions and validation. Use Patito type annotations (`pt.DataFrame[Schema]`, `pt.LazyFrame[Schema]`) whenever a function consumes or returns data that conforms to an existing schema — whether the function is public or private. Don't invent a new schema just to annotate a private helper; if no existing schema fits, use plain `pl.DataFrame` / `pl.LazyFrame`.
 - **Ruff**: 100-char line length, double quotes, Google-style docstrings.
+- **Comments must reflect current state only** — never reference previous iterations of the code,
+  deleted files, or temporary implementation plans (e.g. `docs/temp/`).
+- **MkDocs-compatible constant docs** — document module-level constants with a string literal
+  immediately after the assignment, not with Sphinx-style `#:` comments. This is correct:
+  ```python
+  MY_CONST: Final[str] = "value"
+  """One-line summary.
+
+  Optional further detail.
+  """
+  ```
 - `snake_case` for variables/functions, `PascalCase` for classes, `UPPER_SNAKE_CASE` for constants.
 - All function signatures must have complete type hints including return types.
+- All consts must be marked with the maximally "constant" type. 
+  e.g. `CONST_SEQ: Final[tuple[str, ...]] = ("a", "b")` or `FOO: Final[str] = "bar"`
 - Never relax an existing test to make it pass.
 
 ### Polars Style
--- a/docs/api/ml_core/index.md
+++ b/docs/api/ml_core/index.md
@@ -4,5 +4,4 @@
 
 ::: ml_core.features
 ::: ml_core.base_forecaster
-::: ml_core.cross_validate
 ::: ml_core.metrics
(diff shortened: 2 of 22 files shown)


JackKelly:
And issue to consider (written by Claude):

First, let me restate the root cause precisely, because there's actually a related issue lurking that sharpens the trade-offs.

Root cause

_apply_power_lag is a self-join of the engineered frame against itself (features.py:573-575, _apply_power_lag(engineered_lf, engineered_lf, lag_feat)), keyed on target_time = valid_time - lag_hours. The source of lag values is therefore whatever the engineered base frame is — and that base frame differs by mode:

- Bulk (_join_nwp_bulk_mode): NWP-centric. One row per (ts, nwp_init_time, valid_time, ensemble_member), with observed power broadcast in by a left join on (ts, valid_time). Lag values can only resolve to timestamps that exist as an NWP valid_time.
- Single-run (_join_nwp_single_run): power-centric. One row per observed (ts, valid_time). Lag values resolve against the dense half-hourly observed-power series.

So the lag source is gappy in bulk and dense in single-run. That's the equivalence gap.

The related latent bug (worth surfacing)

Because the bulk frame has the same (ts, member, valid_time) repeated once per overlapping NWP run, the lag self-join fans out: a left row whose target_time lands on a timestamp covered by N runs matches N identical right rows → N fully-identical duplicate output rows. With real daily, overlapping NWP runs over a validation year, any power lag in bulk mode will produce duplicates that PowerForecast.validate will reject (or, where validation isn't called, silently corrupt training). The current tests don't catch it because the only multi-run bulk test requests temperature_2m, not a power lag. Phase 5 (cv_power_forecasts, bulk predict) is where this would bite. This isn't hypothetical — xgboost.yaml uses power_lag_24h/48h/168h/336h.

This matters for the discussion because it means "leave power lags as-is" isn't truly cost-free — it defers a real defect.

Which mode is "correct"?

In production at t0, power_lag_24h for valid_time V should be the real observed power at V−24h (and null when V−24h > t0, which _nullify_leaky_lags already handles via lead time). The single-run/dense computation is the physically correct one. The bulk/gappy computation is the degraded one. So this isn't a symmetric "two valid interpretations differ" situation — bulk is the one that's wrong, and the skew is training (bulk) seeing worse lags than serving (single-run) would.

The options

A — Status quo + scope the equivalence test to weather/time features (what's implemented).
- Pros: minimal; honest; the test is meaningful for exactly the columns that flow through the NWP join. Green now.
- Cons: leaves a genuine train/serve skew for power lags; doesn't surface or fix the fan-out duplication; the "no-skew guarantee" is only partial.

B — Source power lags from the dense observed-power series in both modes (the real fix).
_apply_power_lag already has a separate source_lf parameter — currently both args are engineered_lf. Pass the raw power frame instead, join on (ts, target_time) only (no ensemble_member — observed power has no ensemble dimension; keying on it today is spurious and only "works" because power is identical across members).
- Pros: bulk ≡ single-run for power lags (full no-skew); eliminates the fan-out duplication; more physically correct; the equivalence test can then legitimately include power lags; reuses an existing seam.
- Cons: it's a features.py change, which §5.3 explicitly listed as "stays the same" — so it needs your sign-off and is arguably its own small change (best landed before Phase 5). Bulk-mode feature values change, so some existing feature tests and any cached training data shift. Needs care that the dual-strategy weather lag is untouched (it is — only power lag changes).

C — Make bulk emit the full 6-hourly power_fcst_init_time grid (the §5.7/§6 future refinement).
- This addresses a different skew: bulk currently emits one forecast per daily NWP run, under-sampling the 4×/day production cadence (06/12/18 reuse older NWP with fresher power lags). Orthogonal to the lag-source issue; doesn't fix gappiness or duplication on its own.
- Pros: closes the intraday-cadence skew. Cons: bigger; ~4× bulk dataset; explicitly deferred in the plan.

D — Test-only mitigation (constrain fixtures so lag targets land in one run's window, or compare only the non-null intersection).
- Pros: lets the test "include" power lags with no production change. Cons: tests a happy-path subset, gives false confidence, brittle; the skew + bug remain.

E — Pin the expected difference instead of asserting equivalence.
- A regression test that documents bulk-gappy vs single-dense as deliberate. Pros: honest, cheap, guards against further drift. Cons: enshrines a skew rather than fixing it; wasted if B is the right call.

My read

B is the principled fix and it also clears the latent duplication defect that Phase 5 will otherwise hit — so it has value well beyond the test. A is a fine, honest interim and shouldn't be "reverted"; B simply upgrades it so the equivalence test can then cover power lags too. C is real but separate and belongs as deferred future work. D/E I'd avoid as primary approaches.

If you want B, I'd suggest it as a small standalone change (features.py + the source_lf plumbing + extending the equivalence test to include power_lag_*), landed before Phase 5 rather than folded into the Phase 1 commit.

Let me get your steer on direction.

PR #404, opened 2026-07-21
Stamp git SHA and Delta table versions on MLflow runs (#227)

### LINKED ISSUE #227: Reproducibility: stamp git SHA and Delta table versions on MLflow runs
A run ID today cannot answer "exactly which code and which data produced this model?" — experiments log config and hyperparams but no git commit and no data versions. Stamp the git SHA explicitly (MLflow's auto-detection needs gitpython + repo CWD, neither of which holds in a production container) and log each Delta table's version at read time — Delta time travel makes data versioning one integer per table.

Design: [docs/roadmap/engineering-health.md → "Reproducibility"](https://openclimatefix.github.io/nged-substation-forecast/roadmap/engineering-health/#reproducibility-stamp-git-sha-and-delta-table-versions-on-mlflow-runs)

JackKelly:
Closes #227.

## What

A run ID could not answer "exactly which code and which data produced this model?" — experiments logged config and hyperparams but no git commit and no data versions. This stamps both onto every MLflow run.

New helper `packages/ml_core/src/ml_core/_repro.py`:

- **`get_git_info()`** — `{git_sha, git_dirty}` via `subprocess` (`git rev-parse HEAD`, `git status --porcelain`). Degrades to `"unknown"` outside a git repo (production containers) rather than raising. Stamped explicitly because MLflow's `mlflow.source.git.commit` auto-detection needs gitpython **and** a repo working directory, neither of which holds in a container.
- **`get_delta_versions()`** — each Delta table's `version()` (time travel makes data versioning one integer per table). Missing/unreadable table → `"absent"`, never raising.
- **`provenance_tags(stage, ...)`** — assembles stage-prefixed git + Delta-version tags.

## Where it's stamped

| Asset | Run | Tag prefix | Delta versions stamped |
|---|---|---|---|
| `register_experiment` | parent | `register_` | — (reads no data) |
| `trained_cv_model` | fold | `train_` | power_time_series, nwp_data, eligible_time_series |
| `cv_power_forecasts` | fold | `predict_` | power_time_series, nwp_data |
| `metrics` | fold + parent | `metrics_` | power_forecasts, power_time_series, effective_capacity |

## Two decisions worth a look

1. **Stage-prefixed tag keys.** `trained_cv_model`, `cv_power_forecasts` and `metrics` all write to the *same* fold run, and each may run days apart on a different SHA. A bare `git_sha` tag would clobber — the run would show only whichever asset ran last. Prefixing (`train_git_sha`, `predict_git_sha`, `metrics_git_sha`) keeps all three snapshots. The load-bearing one is `train_*` — it pins the code+data that produced the model.

2. **`storage_options` threaded through** (deviation from the roadmap stub). The `Settings` table paths are `str` and may be remote S3 URIs, so `get_delta_versions` takes `storage_options` and passes it into `DeltaTable(...)`; without this every remote table would report `absent`.

Each stage stamps only the tables it actually reads (e.g. prediction does **not** read `eligible_time_series` — the trained series come from the loaded model — so it isn't stamped there).

## Reconstructing a run

`git checkout {train_git_sha}`, then `pl.scan_delta(power_time_series_path, version=<train_delta_version__power_time_series>)`. Documented in `docs/ml_experimentation/dagster-workflow.md`.

## Tests

- `packages/ml_core/tests/test_repro.py` — git info in a temp repo (clean/dirty), `unknown` outside a repo, Delta version `0` → `1` on append, missing table → `absent`, stage-prefix shape.
- `tests/test_register_experiment_job.py` — extended to assert the parent run carries `register_git_sha` (40-char SHA in-repo).
- Full `uv run pytest` green from the repo root (369 passed, 1 skipped); ruff/format/ty/pymarkdown clean.

## Ship-time triage (done here)

Deleted the "Reproducibility" section from `docs/roadmap/engineering-health.md` and its entry in the page's status banner.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -202,6 +202,19 @@ Each `BaseForecaster` also carries a `feature_engineer: ClassVar[FeatureEngineer
 
 - `snake_case` for variables/functions, `PascalCase` for classes, `UPPER_SNAKE_CASE` for constants.
 - All function signatures must have complete type hints including return types.
+- **Prefer self-documenting type hints over bare containers — a signature is documentation.**
+  Jack strongly prefers expressive signatures and is happy to spend a few extra lines of code to
+  get them, as long as complexity stays low. Whenever you would write `dict[str, str]` (or a bare
+  `str` for a value from a fixed set, or a tuple of positional values), stop and ask whether a more
+  self-documenting type is practical. Reach for: a `Type`-suffixed `Literal` alias for a closed set
+  of string values (`StageType = Literal["register", "train", "predict", "metrics"]`); a named
+  alias for a recurring shape (`MlflowTags = dict[str, str]`) so the intent is stated once and
+  reused; a `TypedDict` for a structured mapping with known keys (e.g. `ObjectStoreOptions`) —
+  taking the `TypedDict` in the signature and widening to a plain dict at the call boundary.
+  Constraining `dict` *keys* to a `Literal` alias (`dict[TableNameType, str]`) is worthwhile for a
+  closed vocabulary and works with bidirectional inference when callers pass dict literals.
+  `packages/ml_core/src/ml_core/_repro.py` is the worked example. Don't force it where no honest
+  stricter type exists — a genuinely heterogeneous or open-ended dict stays `dict[str, str]`.
 - All consts must be marked with the maximally "constant" type.
   e.g. `CONST_SEQ: Final[tuple[str, ...]] = ("a", "b")` or `FOO: Final[str] = "bar"`
 - Never relax an existing test to make it pass.
--- a/docs/ml_experimentation/dagster-workflow.md
+++ b/docs/ml_experimentation/dagster-workflow.md
@@ -261,6 +261,26 @@ FastAPI server.
 Full MLflow (which bundles the web server) is in the `dev` dependency group, so `uv sync` installs
 it; production runs use `mlflow-skinny` (the client, without the server).
 
+## Reproducibility: reconstructing a run from its stamped provenance
+
+Every run carries the code and data provenance needed to answer "exactly which code and which data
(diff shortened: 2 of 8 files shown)


JackKelly:
## Adversarial review + fixes (commit ca872f9)

Ran an adversarial review of the diff. It confirmed the load-bearing design is sound — **clobber prevention** (stage prefixes fully disjoint across parent and fold runs), **per-stage table accuracy** (each stage stamps exactly the Delta tables it reads; prediction correctly omits `eligible_time_series`), **`storage_options` plumbing**, **MLflow tag-key limits**, **idempotency** (`set_tags` overwrites cleanly on retry), and the PEP 758 syntax — all clean.

It found real holes in `get_git_info`'s "never raises" contract, now fixed:

| # | Severity | Issue | Fix |
|---|---|---|---|
| 1 | Medium | Narrow `except` let `UnicodeDecodeError` (non-UTF-8 filename + `quotePath=false`) and `TimeoutExpired` escape | broadened to `except Exception` + `errors="replace"` on decode |
| 2 | Medium | No subprocess timeout → a stalled `.git` (NFS/stale lock) hangs the asset forever | 5s cap per git call |
| 4 | Low | A failing `git status` discarded an already-captured SHA | split the two calls; only `git_dirty` degrades to `unknown` |
| 3 | Medium | `_GIT_CWD` walks upward, so a non-editable install nested in an unrelated repo could report a wrong SHA | docstring caveat (a confidently-wrong SHA is worse than `unknown`; trust the SHA only for editable/workspace installs) |
| 5 | Nit | `storage_options or {}` vs `storage_options` inconsistency | unified |

Added a test that `get_git_info` never raises on a nonexistent `cwd` (exercises a different exception path than the out-of-repo case). Full suite still green.

**Not fixed (accepted):** the `metadata_path` parquet is read by all three stages but isn't version-stamped — it's parquet, not Delta, so it has no version. Provenance for metadata is simply absent by nature.

Issue #424, opened 2026-08-03
Implement asset check for power forecasts


JackKelly:
To catch if the forecast asset "succeeds" (it runs and doesn't raise any exceptions) but writes invalid forecasts to disk (or no forecasts at all).

This isn't about measuring the _skill_ of our live forecasts. Just check that the rows exist and have valid data.

This is related to (but distinct from) measuring the _skill_ of our live forecasts, which is tracked in #224

---

## Notes from the inherent-stability design work (#429)

**Ship this as `severity=WARN`, `blocking=False`**, matching `power_data_is_fresh` and `nwp_has_no_unexpected_nulls`. There is currently no ERROR-severity check anywhere in the repo, and a blocking check here would contradict the design principle that a partition fails only when there is genuinely no useful data.

**Also report missed NWP runs at forecast time — not raw NWP age.** Every production asset has a check except `live_forecasts` — the one NGED actually consumes. Two failure modes need surfacing:

- NWP more than ~15 days stale currently causes a **hard failure**: a weeks-old run no longer covers the forecast window, its rows join to a null `ensemble_member`, get filtered out, and the asset raises on zero rows.
- NWP between 0 and ~15 days stale **degrades silently** — the forecast is built from an increasingly ancient run, with `nwp_init_time` on the row but no warning anywhere.

**Count missed runs, not hours.** We ingest one ECMWF run per day (00Z, downloaded at 08:30 UTC by `ecmwf_ens_schedule`) and forecast at 00/06/12/18, so *healthy* NWP age at forecast time is 12–30 h — 24 h at the 00:00 slot, 30 h at 06:00 (just before the day's download lands), 12 h at 12:00, 18 h at 18:00. An absolute-age threshold low enough to catch a real outage would fire on two of the four slots every day. The health signal is how many daily runs are missing between the freshest run on disk and the freshest that should exist by now, which is zero in every healthy slot.

[...]

Issue #423, opened 2026-08-02
Tag Dagster assets with tags "R&D" or "production"


JackKelly:
The aim is to make it easier to see which assets are required on production, so the person operating the production system doesn't have to worry about the "noise" of the assets that are concerned with cross validation etc.

JackKelly:
Note from the [inherent-stability](https://openclimatefix.github.io/nged-substation-forecast/design-philosophy/inherent-stability/#rd-fails-the-other-way) work: **this tagging is the mechanism behind the fail-fast / fail-operational asymmetry.**

Production and R&D have opposite costs of being wrong. In production the cost of *no output* is high (a user is waiting) and the cost of a flagged, degraded output is moderate — so the correct posture is fail-operational. In model R&D the cost of no output is nil (rerun it) and the cost of a quietly-degraded output is high (it silently poisons a model and every comparison built on it) — so the correct posture is fail-fast.

That asymmetry is already implemented at the telemetry layer (the Sentry failure hook is attached to the three scheduled production jobs only), but it is implicit in the code rather than expressed. The R&D/production tag introduced by this issue is what makes it explicit and queryable — worth stating in the issue’s rationale, because it is a stronger reason for the tag than observability alone.

Issue #420, opened 2026-07-24
Silence warnings for known "dead" time series?


JackKelly:
James says: "It turns out that PV site in TimeSeriesInstance 23 has been off since the 4/7/26.... and the monitor for site 33 is just broken."

So maybe we should have a way to express to the code "we know about these broken time series so you don't have to warn me very day about these (but maybe still send a message to say "ignoring time series ids 23 and 33", so I remember that we've turned that off). Please alert me if/when they re-appear. And definitely warn me if any other time series go bad."

JackKelly:
Context: this is the **warning-fatigue** counterweight to the `power_forecast_warnings` work (#439 Phase 1, #441 Phase 2) and the `asset_health_history` table (#442). A warning channel that fires constantly for series that are dead rather than broken trains the operator to ignore it, which defeats the [provider channel](https://openclimatefix.github.io/nged-substation-forecast/design-philosophy/inherent-stability/#three-audiences-three-channels) entirely. Worth sequencing downstream of those three so the silencing rules are written against the real warning vocabulary rather than a guess at it.

JackKelly:
Claude's questions:

> where should the known-dead list live? Options: a YAML file in conf/, an env var in .env, or a column in the metadata parquet. The issue also asks for three behaviours (a quiet "ignoring ids 23, 33" note, an alert if they come back, and normal warnings for everything else) — I'd default to a YAML file in conf/ since it's reviewable in git, but say if you'd rather it be config-not-code.

I think the list of dead timeseries should live with the code (because a time series is dead, no matter if our code is running on a laptop or AWS). So maybe YAML in `conf/`?

JackKelly:
Given our aim to build a system that could, in principle, be operated by NGED, without them writing any code, I think the idea solution to this issue would actually be to be able to add and remove time_series_ids to a "known late" list and maybe a separate "known_missing". I think the ideal behaviour would be something like:

- If Dagster finds that a time_series_id is late or missing, then it warns the operator (as happens now), and somehow makes it as easy as possible in the UI to let the operator add one or more of these time_series_ids to their "ignore" list
- Then, if Dagster subsequently finds that a "known missing" or "known late" ID _appears_, then it sends a message to Sentry (making it clear that this is a "happy message"!), and automatically removes it from the "known missing" or "known late" list

JackKelly:
Related issue:

-  #523

Issue #374, opened 2026-07-20
Add more data-validation functions


AUdaltsova:

JackKelly:
A concrete shape for this issue, from the [inherent-stability](https://openclimatefix.github.io/nged-substation-forecast/design-philosophy/inherent-stability/#missing-versus-wrong) work — the **missing versus wrong** distinction, which is exactly what "more data-validation functions" is for:

- **Absent or stale input** → always produce a forecast. Degrade, widen the bands, declare it.
- **Detectably wrong input** → do not consume it. Treat it as *missing*, which routes it back into the always-output path.

The second half is the gap. A stuck meter reporting 2.1 MW for 52 hours is not missing data; it is actively misleading, and a lag-feature model will propagate it happily. The manual heuristic has the identical vulnerability. [Data quality](https://openclimatefix.github.io/nged-substation-forecast/background/data-quality/) already documents false zeros, stuck values and genuine gaps as separate phenomena with real plots, so the evidence base exists — what is missing is the detectors that turn "wrong" into "missing" before the data reaches a model.

Note the posture this implies, which is deliberate and the opposite of Postel’s law: **liberal about missing inputs, strict about malformed ones**. Validation functions added here should reject or null the malformed value rather than repair it.

PR #428, opened 2026-08-04
Document the inherent-stability principle, and open the workstream it implies

### LINKED ISSUE #429: Document the inherent-stability design principle
Write down the design principle the pipeline is already built around, and make the engineering claims that follow from it falsifiable.

**The principle.** The service should degrade gracefully and legibly rather than fail, and should communicate that degradation *in the answer itself* rather than through a side channel. A Dagster asset partition should fail only when there is genuinely no useful data; anything short of that produces a forecast, with a warning and a wider uncertainty band. The manual heuristic consumes no NWP at all and survives a power-data outage, which makes it a measurable floor: we should only fail when we can no longer beat it.

## Deliverables

- **`docs/architecture/inherent-stability.md`** — the principle, a failure-mode table (what breaks / what the system does / whether a human is paged), the rules as a numbered list, the mechanisms, and what this deliberately is not (Postel's law; the R&D-versus-production asymmetry).
- **`docs/engineering-hypotheses.md`** — H1 (manual attention only for upstream format changes), H2 (hundreds of experiments a month), H3 (one-click promotion *and* rollback), each with a test, a threshold and a measurement source. All three already exist as prose in `background/requirements.md`; this extracts and elevates them rather than restating them.
- **Edits to ~14 existing pages**, including a home for both new pages in `documentation-guide.md`'s four-place model, a cross-link from `architecture/overview.md`, and a fourth bullet under `requirements.md` § "Uptime: lenient by design" — its three bounded-damage arguments all assume the outage is *our compute stopping*, which does not cover an extended NWP outage.
- **A short entry in `CLAUDE.md`**, so the principle reaches future Claude Code sessions.

## Follow-on work this issue does not cover

The design work identified a set of implementation items that will get their own issues, tagged with an `inherent-stability` label: the failure-scenario suite and its leaderboard scoring, `power_forecast_warnings` (which has no issue at all despite being a contractual v1.0 delivery table), degradation-conditional interval calibration, and an intervention log to make H1 measurable.

It also found three divergences in `live_forecasts`, where an extended NWP outage causes a hard failure rather than a degraded forecast, and where moderate NWP staleness degrades silently. Those interact with #424 — the check must land before the degradation change, or a loud failure becomes a silent one.

Plan and full analysis: #428.

JackKelly:
Documents **inherent stability** — the principle that the forecasting service should degrade gracefully and legibly rather than fail, and should communicate that degradation *in the answer itself* rather than through a side channel — and turns it into permanent docs plus a tracked workstream.

Closes #429.

> This PR started as `plans/inherent-stability-plan.md` (a design document for review). All five of its open decisions have now been answered, so the plan has been executed and deleted. The parts of it that do not belong in `docs/` are preserved at the bottom of this description.

## Two new pages

**`docs/design-philosophy/inherent-stability.md`** — the durable design rationale:

- The principle, the incumbent-as-floor argument, and why the analogy is self-centring steering rather than a nuclear reactor (in a reactor *safe means off*, which is the opposite of the thesis).
- The **degradation ladder** — five rungs from "everything fresh" to "nothing at all", the last still bounded by clear-sky physics.
- A **failure-mode table** stating honestly what the code does *today* versus what it should do, and whether anyone is paged.
- **Ten numbered rules** for production code — the part future contributors (and Claude sessions) will actually act on.
- The mechanisms, and what the principle deliberately is **not**: Postel's law (disowned), blast radius (a different axis), and the R&D asymmetry, which fails the other way on purpose.

**`docs/design-philosophy/engineering-hypotheses.md`** — six falsifiable claims with thresholds and resolution points (H1 operability, H2 experiment throughput, H3 one-command promotion/rollback, H4 cost, H5 operability by a non-expert, H6 scale without redesign), plus the intervention log that T1.1 is scored from. NIA funding is for transferable learning including negative results, so a pre-registered number we then miss is a publishable finding.

## Fifteen edits to existing pages

The principle only works if it reaches the pages where the relevant decision is made:

| Page | What was added |
|---|---|
| `xgboost-improvements.md` | The **NaN default-direction limit** — XGBoost handles the missingness it saw *during training*, so a model trained with NWP and run without it is not thereby weather-blind |
| `capacity-estimation.md` | **Missingness robustness as a head-to-head judging criterion** — DP's structural advantage here should count in the v0.7 contest |
| `techniques/differentiable-physics.md` | Graceful degradation as a fourth reason for DP; a physical forward model has a defined output for any input state |
| `techniques/encoders.md` | Token removal, not zero-fill (zero is a real physical value); and why random dropout trains for the wrong world |
| `metrics-and-leaderboard.md` | **Scoring under failure scenarios** — the suite, how it is scored, and `nged_incumbent` as the acceptance criterion |
| `engineering-health.md` | Degradation smoke-tests in the scientific-rigor suite |
| `delivery-tables.md` | `warning_source` on Table 2, and degradation-conditional band widening on Table 1 |
| `live_service/operations.md` | **Two new runbooks**: rolling back to the previous champion, and degraded input data |
| `background/requirements.md` | The case "Uptime: lenient by design" does *not* cover — an extended NWP outage keeps compute running while the last good forecast ages out |
| `documentation-guide.md` | A home for the hypotheses page in both taxonomy tables |
| `why-dagster-not-airflow.md` | The asset-checks gap is architectural, not cosmetic; and a new cost against Option B |
| `architecture/overview.md`, `docs/index.md`, `roadmap/index.md` | Cross-links and milestone bullets |
| `CLAUDE.md` | A short rules entry, since a docs page alone will not reliably reach future Claude sessions |

## The workstream

Fourteen new issues, all labelled `inherent-stability` (`gh issue list --label inherent-stability`), attached to their milestone epics with `blocked by` links, and added to the OCF board:

| Milestone | Issues |
|---|---|
| **v0.2** | #435 intervention log · #424 missed-NWP-runs check (existing; extended) |
| **v0.3** | #437 failure-scenario suite · #436 degradation smoke-tests · #438 leaderboard scenario scoring · #440 one-command rollback · #448 promotion/rollback runbooks · #447 degraded-data runbook |
| **v0.4** | #439 warnings Phase 1 · #441 warnings Phase 2 · #442 `asset_health_history` |
| **v0.5** | #443 conformal calibration per regime · #445 weather-blind training · #444 clear-sky floor · #446 `live_forecasts` degrades rather than raises |

Notes (not duplicates) were added to #362, #423, #141, #157, #158, #374, #147, #168, #420 and #161.

The sequencing argument throughout: **the evaluation machinery must precede the model experiments it is meant to judge**, which is why v0.3 carries the scenario suite and v0.5 carries everything that depends on it.

---

## Preserved from the deleted plan

**The code audit (2026-08-04).** The audit found the pipeline already largely in line with the principle: `promoted_model` copies the champion to local disk so inference makes no MLflow call; `live_forecasts` selects "the freshest NWP run present as of the forecast time", a relative query with no fallback branch to get wrong; both asset checks are non-blocking WARN; there is no `ERROR`-severity check anywhere in the repo; Sentry's log-to-event capture is deliberately disabled so a fail-open design does not flood it; the failure hook is attached to the three scheduled jobs only; and `report_power_freshness` never raises. Three divergences were found, all in `live_forecasts`, and all are now issue #446 — with the ordering constraint that #424's check must land first, or a loud failure becomes a silent one.

**Effort and sequencing.** The expensive items are not deferred by choice; they are blocked by the v0.5 quantile pipeline. The items that must be early are the cheap ones (the scenario suite is pure transforms over an `AllFeatures` frame — the vocabulary design is the work, the code is small; leaderboard scenario scoring costs N× *predict*, not N× train). So there was no real trade-off between groundwork now and implementation later.

**Two items were cut or deferred on review.** Cost-per-experiment instrumentation was cut outright — training runs on laptops and costs AWS nothing, so H4's measurement is reading the monthly bill. And a `docs/techniques/conformal-prediction.md` explainer is deliberately *not* written yet: it is a hard precondition on #443 rather than work for this PR, because the term appears nowhere in the docs today and writing the page now would mean maintaining an explainer for a technique nothing yet uses.

**The `power_forecasts` schema decision.** The run-level metadata stays denormalised rather than moving behind a `power_fcst_id` into a dimension table: those columns are run-constant and so nearly free once dictionary- and run-length-encoded, a surrogate key would be the table's only incompressible column on a table that passes 2³² rows at V2, a join defeats the Delta predicate pushdown the storage policy is built around, and it would break the `replaceWhere` slot-overwrite path `live_forecasts` relies on. The rule that generalises: run-constant metadata belongs on the fact table; genuinely per-row-varying metadata deserves a second look. `scenario` therefore becomes a plain `String` column heading `POWER_FORECASTS_SORT_COLS`, not a third partition column — details in #438.

## Verification

- `uv run mkdocs build --strict` — clean.
- `uv run pymarkdown scan -r docs README.md CLAUDE.md ...` — clean.
- Every relative link and heading anchor across all of `docs/` resolves (checked by rendering each page's headings through Python-Markdown's slugify and matching every `](...#anchor)`).
- Rendered HTML spot-checked for the two known Python-Markdown list traps: nested sub-bullets use 4-space indents, and sibling bullets after an indented continuation have a blank line before them.


JackKelly:
## Audit pass: checked the plan against the code, corrected five claims

### The WARN/ERROR and Sentry discipline is already largely in line with the plan

Audited `defs/checks.py`, `defs/assets.py`, `defs/production_assets.py`, `defs/schedules.py`, `definitions.py` and `_sentry.py`. Findings now recorded as §4.1:

- **Both asset checks are WARN and non-blocking**, and **no ERROR-severity check exists anywhere in the repo**.
- **Sentry's log-to-event capture is deliberately disabled** (`LoggingIntegration(event_level=None)`) — subtle and exactly right, since otherwise every `ERROR` log becomes an event and a fail-open design would flood Sentry with events for conditions it deliberately tolerates.
- **The failure hook is attached to the three scheduled jobs only** — the R&D/production asymmetry is already implemented at the telemetry layer.
- **`report_power_freshness` never raises**, with a docstring recognising that a bug in the *warning* path could silently convert fail-open into fail-closed.

### Three divergences, all in `live_forecasts`

1. **NWP >~15 days stale → hard failure.** A weeks-old run no longer covers the forecast window, rows join to a null `ensemble_member`, get filtered at `production_assets.py:280`, and the asset raises on `forecasts.height == 0` at `:286`. Rung 2 of the degradation ladder, and NGED gets nothing.
2. **NWP 0–15 days stale → silent degradation.** No check, no Sentry event, no widened bands. Arguably worse than the hard failure.
3. **`select_nwp_init_time` raises** when nothing qualifies (`_production_helpers.py:71`) — narrower, but same class, and it bites `replay` backfills.

The raise on empty `trained_ids` is *correctly* fail-fast and stays — an empty model is a promotion bug, not a data outage.

**Load-bearing ordering:** fixing 1 without fixing 2 converts a loud failure into a silent one, so the check must land first.

### Corrections from re-checking against code and GitHub

| Was | Actually |
|---|---|
| "New issue: a `live_forecasts` asset check" | **Already exists as #424.** Extend it (add NWP-age reporting, severity decision, epic, project fields) rather than duplicate |
| "New issue: clear-sky physical envelope" | **Already a deliverable of #168**, with substantial physics in `docs/techniques/differentiable-physics.md`. Only the *floor* framing is new |
| T1.2 scores against `nged_incumbent` | `nged_incumbent` is **designed but not implemented** — no Python match. **Blocked on #147** |
| "Milestone" column | The repo has **zero GitHub milestones**. Every item attaches as a sub-issue of a parent epic |
| `EVALUATION_SCOPES` | `EvalScopeType` is a deliberately narrower `Literal` with a documented expansion pattern; a new scope must follow it |

Also: `requirements.md` L98 → L94, and the ~£25–35/month figure covers the whole v1 stack, not just the control-plane box.

### Effect on scope

Items 13 and 14 are **production-correctness fixes, not quality work** — divergence 1 is a live hard-failure mode. So they sit *inside* "live service first" rather than against it, and the scope widening is smaller than the original plan implied.

JackKelly:
## Docs review: all of `docs/` (excluding generated `api/`)

Now §7.7 in the plan. **Headline: no doc currently overclaims.** I searched specifically for pages asserting degradation behaviour the code doesn't deliver (§4.1). None found — the two closest (`production-deployment.md:42`, `requirements.md:152`) both argue that a *missed slot* degrades freshness rather than cutting NGED off, which is true, because delivery is decoupled from compute. So this work *adds* to the docs rather than correcting them. With one exception, finding 3.

### Finding 1 — neither new page has a home in the repo's own documentation model

`documentation-guide.md` defines four places with "deliberately non-overlapping jobs".

- **`engineering-hypotheses.md` fits none of them.** Not unbuilt design, not an explanation of built code, not NGED-derived background. It's a fifth kind of content: our own falsifiable claims. The guide needs a new row in both its tables, or the page violates the documented structure on arrival.
- **`architecture/inherent-stability.md` spans the architecture/roadmap line**, because the philosophy is partly built (WARN checks, static NWP selection) and partly not (widening bands, scenario suite). Resolution: principle + built mechanisms in `architecture/`; unbuilt mechanisms stay in their roadmap pages and are **linked, not copied** — otherwise the page becomes a roadmap mirror and rots.

### Finding 2 — `operations.md` is missing the two runbooks this philosophy depends on

Its headings: Prerequisites → Pick a champion → Materialise `promoted_model` → Let the schedule run → Inspecting a live forecast → Backfilling a missed slot. **Zero occurrences of "stale", "outage", or "missing".**

- **No runbook for degraded input data.** H1's threshold is literally "recovery next business day, via runbook" — unmeasurable without one.
- **No rollback procedure.** H3's second half showing up as a docs gap as well as a code gap.

New plan items **15** and **16**.

### Finding 3 — `requirements.md` § "Uptime: lenient by design" has a hole mirroring the code's

It bounds outage damage with three arguments (14-day horizon, delivery decoupled from compute, legacy fallback) — all of which assume the outage is *our compute stopping*. An extended NWP outage is a different shape: compute keeps running, forecasts keep being attempted and hard-fail, and the last good forecast ages out too. Needs a fourth bullet or an explicit caveat.

### Finding 4 — three pages already carry this content; link, don't restate

| Page | Already says |
|---|---|
| `background/data-quality.md` | Separate **False zeros** / **Stuck values** / **Missing data** sections — the missing-vs-wrong distinction, already documented empirically with plots |
| `architecture/production-deployment.md` (~L80) | The WARN-not-failure reasoning almost verbatim, *and* the inside/outside monitoring complementarity §3.5 calls load-bearing |
| `ml_experimentation/mlops-approach.md` | "Nothing gets rewritten on the way to production" — H3's substance |

### Finding 5 — `architecture/overview.md` is currently billed as the design-philosophy page

`docs/index.md` describes it as "design philosophy, technical components, and data flow". Without a cross-link, the two pages compete for the same job.

---

Docs-edit list in §7.1 grew from 7 pages to 13. Item count is now 17.

PR #430, opened 2026-08-05
Docs: could this codebase forecast another country? (India thought experiment)

JackKelly:
## Why

OCF is pitching for an innovation project in India: ~100,000 secondary substations reporting every 15 minutes, forecasting net demand *and* PV per substation, where the PV is entirely unmetered and there are no installed-capacity priors. Jack asked what adapting this codebase would take. This PR writes that assessment down so the reasoning stays auditable and shareable.

**This is explicitly not planned work.** There is no GitHub issue, no roadmap entry, and no intention to refactor for portability unless we win the bid. The page says so in a status banner at the top *and* in a closing section, so a reader landing mid-page can't miss it.

## Where it lives, and why not `docs/roadmap/`

`docs/architecture/adapting-to-another-geography.md`, titled **"Could this codebase forecast another country?"** — nav entry under Architecture, immediately after "Why Dagster, not Airflow?".

Per the [Documentation Guide](https://openclimatefix.github.io/nged-substation-forecast/documentation-guide/), `docs/roadmap/` is design for work we intend to build — every substantial page has an epic, and pages are *deleted* when their items ship. A speculative assessment there would falsely signal intent and would never get cleaned up, because it never ships. `architecture/` is durable and already hosts exactly this genre: [Why Dagster, not Airflow?](https://openclimatefix.github.io/nged-substation-forecast/architecture/why-dagster-not-airflow/) is a "could we port to X?" analysis recorded to keep the reasoning auditable, ending in "not now" plus the triggers that would change that. This page mirrors its shape.

The page opens with a **summary written for a non-coder** who understands the energy system — no file paths, no class names, no Polars/Dagster/Delta jargon. Technical detail follows below it.

## What the assessment concluded

Three quite different pieces:

- **The geography swap is cheap** (~4–6 weeks). More is already neutral than expected: the ECMWF catalogue we use is *global*, so India needs no new weather source; the download derives its bbox from the H3 grid at runtime; `geo/h3.py` takes any boundary; and lags/rolling windows are already durations rather than counts of half-hour periods. What is British is a thin layer in `contracts` — lat/lon bounds that hard-fail on Indian coordinates, four enums, MW thresholds — plus one GB outline and one `"Europe/London"` literal.
- **Scale is the real engineering** (~3–4 months). 100k × 15-min is ~80× the V2 design point. One booster per series becomes a non-starter, forcing the global model already planned as #104 ("the stepping stone to V2 scale"); `power_time_series` partitioning by `time_series_id` would mean 100k Hive directories; a full-ensemble run is ~6.9B rows, ~18 TB/year.
- **Unmetered PV disaggregation is the same research bet we are already making.** `UniversalSolarFleetNode` and the convex dictionary baseline both work without a capacity register. India is harder in one way (no metered PV to anchor against) and easier in two (100k sites vastly better constrains the shared-basis design than 32; 15-min separates solar from load better than 30-min). Flagged honestly: load shedding, diesel gensets and agricultural pumping all break the smooth-weather-driven-demand assumption.

Recommended structure if we ever did it: **one monorepo, not a fork** — a fork means doing the disaggregation research twice.

## One real bug found on the way

`local_utc_offset` computes `(base_utc_offset + dst_offset).dt.total_seconds() // 3600` cast to `Int8` ([`tabular_feature_engineer.py:358`](https://github.com/openclimatefix/nged-substation-forecast/blob/main/packages/ml_core/src/ml_core/features/tabular_feature_engineer.py)). Britain only has whole-hour offsets so this is correct today, but it would silently truncate IST (UTC+5:30) to +5. **Not fixed here** — it's out of scope for a docs PR and harmless while we only forecast Britain. Happy to open an issue if you want it tracked.

## Verification

- `uv run pymarkdown scan -r docs README.md CLAUDE.md metadata/README.md packages/*/README.md` — clean.
- `uv run mkdocs build --strict` — passes, so every cross-page anchor resolves (the page links into `overview.md`, `disaggregation.md`, `differentiable-physics.md`, `capacity-estimation.md` and `xgboost-improvements.md`).
- Inspected the rendered HTML rather than trusting the linter, per the Python-Markdown list-interruption gotcha in CLAUDE.md: 25 `<li>`, 3 tables, 1 blockquote, zero leaked bullet markers in paragraphs.
- All pre-commit hooks passed.

All file:line references in the page were re-verified against `main` at `737fb86`, and the page is dated accordingly.

### DIFF
--- a/docs/architecture/adapting-to-another-geography.md
+++ b/docs/architecture/adapting-to-another-geography.md
@@ -33,13 +33,13 @@ The assessment assumed the following brief:
 What it would take to point this system at that brief splits into three quite different pieces,
 and they are worth keeping apart because only one of them is genuinely hard.
 
-**Moving to a different country is easy.**  The weather data we already use is a
-global product, so India needs no new weather source at all. The parts of the system that handle
-spatial gridding, storage, and feature building never learned that they were in Britain. What *is*
+**Moving to a different country is easy.**  The weather data we already use (ECMWF ENS from
+Dynamical) is a global product, so India needs no new weather source at all. The parts of the system
+that handle spatial gridding, storage, and feature building are geography-agnostic. What *is*
 British is a thin, well-contained layer: a map outline, a permitted latitude and longitude range,
-the names of British substation categories, and the assumption that readings arrive every half
-hour rather than every 15 minutes. Turning those from hard-coded facts into settings is a few
-weeks of work, not a rewrite.
+the names of British substation categories, and the assumption that readings arrive every half hour
+rather than every 15 minutes. Turning those from hard-coded facts into settings is a few weeks of
+work, not a rewrite.
 
 **Handling 100,000 substations is a real engineering project, but an ordinary one.** It is 40 times
 as many substations as we are currently building towards for NGED, and — because the readings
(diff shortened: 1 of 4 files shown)


JackKelly:
## Adversarial review pass — `0219aba`

Ran a fresh-context subagent tasked with *falsifying* the page. It could not break a single code citation (every file path and line number checked out) nor any of the scale arithmetic (6.9B rows/run, 10 trillion/year, 18 TB, 80×, 78×, 15.7× land area). It did find real problems, all now fixed.

**Three factual errors:**

- "only two modules import `nged_data`" — it is three (`checks.py`, `assets.py`, `cv_assets.py`; the other grep hits are the `nged_data_path` *setting*).
- The quantile mitigation said "roughly 5×". `DELIVERY_QUANTILES` has **thirteen** levels and Representation 2 is one row per `valid_time`, so it is a **51× row reduction** and ~4× fewer stored values. Now stated correctly and linked to [delivery-tables.md](https://openclimatefix.github.io/nged-substation-forecast/roadmap/delivery-tables/#representation-2-percentiles).
- `train()` does not write `.ubj` files — `save()` does.

**Overclaims a domain expert would have flagged in a bid:**

- **"no metered solar in India" was simply wrong.** India has ~100 GW installed, overwhelmingly utility-scale with generation metering. The *brief* offers no in-dataset anchor — a different and defensible claim. Rewritten, and now notes the external fleet as a candidate anchor.
- **Two India-specific physics problems were missing and attack the method directly.** *Soiling*: Indian rooftop PV loses 15–30% seasonally to dust and recovers after monsoon rain, but `UniversalSolarFleetNode` represents capacity as `softplus(increments).cumsum()` — non-decreasing **by construction** — so a reversible loss is structurally inexpressible. In GB the mechanism that absorbs this is effective-capacity tracking, which is scoped to *metered* generators — the very anchor India lacks, so the two gaps compound. *Aerosol*: the Indo-Gangetic Plain has among the world's highest AOD, and capacity is inferred *from* irradiance, so irradiance bias becomes capacity bias. I also found that CM SAF SARAH-3 — our planned high-resolution irradiance source — is a Meteosat prime-disc product that **does not reach India**.
- Agricultural pumping moved from confounder to *opportunity*: Indian agricultural feeders are largely segregated on a published supply schedule, so that load is partly known in advance. Stronger and more credible for a bid.
- "strictly better" → "clearly the favourable choice" (the roadmap lists hard prerequisites for the global model). Effort range now matches its own table: four to five and a half months.

**The `local_utc_offset` finding was overbilled.** In a single-timezone deployment the feature is *constant*, so mapping +5:30 → 5 destroys no information the model could use. The real failure mode is **collision** (India +5:30 and Nepal +5:45 both give `5`) plus illegibility. Also `//` floors rather than truncates, so −3:30 → −4. Page and #431 both updated.

**Omissions added:** the dashboard's second hard-coded `Europe/London`; the private `_H3_RESOLUTION` import that parks the NWP grid resolution inside the ingest package; the ECMWF archive starting only at 2024-04-01 with GHI and no direct component; and — the best catch — **6.9B rows/run exceeds the 2³² cap on its own**, so chunked writes are a precondition, not an optimisation.

I also separated two multipliers the page had conflated: **40×** on series count (what forces the global model) versus **80×** on forecast-row volume (what drives storage).

**Placement was the sharpest finding.** `documentation-guide.md` defines the `architecture/` group as "what is already built", so the page contradicted the taxonomy it was filed under — and an AI tool handed `docs/` would read `RegionProfile` and `PowerIngest` as intended design. Rather than move it (roadmap/ is worse: it implies intent and gets deleted on ship), I added a row to the "Which place do I use?" table admitting the "assessment of work we decided **not** to do" genre, and put "(not planned)" in the nav entry.

`pymarkdown` clean, `mkdocs build --strict` passes, rendered HTML re-checked (27 `<li>`, 3 tables, zero leaked bullet markers).

Issue #324, opened 2026-07-15
Replace polling schedules with Dagster sensors (for fetching fresh NGED & Dynamical data) 


JackKelly:
## Problem

Two of the live schedules launch Fargate tasks that mostly do nothing, because they poll for data by *running the job* rather than by *checking first*:

- **`power_time_series_and_metadata_schedule` (hourly)**: NGED publishes new JSON files only roughly every 5 hours, so ~19 of every 24 hourly runs boot a 4 vCPU / 16 GB Fargate task, pull the image, list NGED's bucket, find nothing new, and exit. That's ~720 tasks/month, most of them no-ops (~$5–10/month, and a lot of pointless task churn in the ECS/Dagster history).
- **`ecmwf_ens_schedule` (daily)**: when the 00Z run isn't published yet, `NwpRunNotYetAvailable` → `RetryRequested(seconds_to_wait=1800, max_retries=8)` retries *inside the same run*, so the 16 GB Fargate task sits idle for up to 4 hours on a late-publication day, billed the whole time.

## Fix: Dagster sensors

Sensor evaluation functions run **on the daemon** — i.e. on the always-on control-plane box — while only genuine *runs* go through the `EcsRunLauncher`. So sensors give us "cheap detection on the box, real work on Fargate" natively, with no custom run launcher:

1. **NGED telemetry sensor** (replaces the hourly schedule): every few minutes (`minimum_interval_seconds` ~300–600), cheaply list NGED's bucket (`list_timeseries_json_files` + a cursor over file listing, or reuse `select_new_rows`'s freshness check) and fire `power_time_series_and_metadata_job` only when there are genuinely new files. Fargate then launches only for the ~5 real ingests/day.
2. **NWP publication sensor** (replaces or augments the daily schedule): poll Dynamical for the day's 00Z run from the box and fire `ecmwf_ens_job` once it's published, so the retry loop (and its idle Fargate hours) can be dropped or kept only as a thin safety net.

## Explicitly out of scope: ordering `live_forecasts` against the ingest jobs

An earlier version of this issue proposed a **run-status sensor** that would fire `live_forecasts_job` only after that slot's ingest had actually succeeded, to "make the ordering rigorous". **Do not do this.**

[...]

PR #452, opened 2026-08-07
Add principle 14: couple through data at rest, not run status

JackKelly:
**Stacked on #451** — targets `intervention-log`, so review that one first. The diff shown here is just this change.

## The idea

Our production jobs don't really depend on each other. `live_forecasts` doesn't care whether `ecmwf_ens` succeeded in the last 24 hours, or whether this hour's telemetry pull ran — it selects the freshest NWP run genuinely present as of its own init time and stamps the row with how stale that was. A failed ingest makes the 06:00 forecast slightly staler. It cannot make it late, and it cannot make it absent.

The common alternative is a chain of scheduled jobs — A at 06:00, B at 06:15, C at 06:30 — where B's real input is *the event of A having run*. That design can't distinguish "A failed" from "A is still running" from "A had nothing to do", so one bad morning takes out everything downstream for the rest of the day.

## Why this needed writing down

The property was already implemented, and already argued for in one place — [`f001982`](https://github.com/openclimatefix/nged-substation-forecast/commit/f001982) dropped the `promoted_model` lineage edge from `live_forecasts` for precisely this reason. But the docs never stated the general rule, and **two places framed it as a defect awaiting a fix**:

- The `power_time_series_and_metadata_schedule` docstring called the `:55` offset "a cheap mitigation, not a guarantee" and carried a TODO proposing a run-status sensor that would fire `live_forecasts_job` *only once the ingest had succeeded*.
- [why-dagster-not-airflow.md](https://openclimatefix.github.io/nged-substation-forecast/architecture/why-dagster-not-airflow/) listed "Sensors / run-status coordination" against "ingest → forecast ordering" as planned work.

Both describe a gate, and a gate turns one failed upstream run into a missing forecast — the exact failure mode [Inherent Stability](https://openclimatefix.github.io/nged-substation-forecast/design-philosophy/inherent-stability/) exists to prevent. An implementer following that TODO would have removed a load-bearing stability property while believing they were hardening it.

`live_forecasts` wants to run at 00/06/12/18 regardless, so there is no acceleration to be gained there anyway. Sensors are still wanted on the **ingest** assets, where they avoid booting a Fargate task just to discover there's nothing new to fetch — which is what #324 is actually for. The comparison-table row now says that instead.

## Changes

- **New principle 14** in [design-principles.md](https://openclimatefix.github.io/nged-substation-forecast/design-philosophy/design-principles/), appended rather than inserted so no existing "principle N" citation moves.
- **New rule 11** in [inherent-stability.md](https://openclimatefix.github.io/nged-substation-forecast/design-philosophy/inherent-stability/), marked as restating a principle; preamble updated from "three of the ten" to "four of the eleven", and principle 1's "ten rules" link with it.
- **`schedules.py`** — TODO removed; both docstrings now say the decoupling is deliberate and why.
- **operations.md** and **why-dagster-not-airflow.md** — reworded off the "not a guarantee / fix still to explore" framing.

## The asymmetry worth keeping in view

The *lineage* edge stays. Dagster still knows `live_forecasts` depends on `ecmwf_ens` and `power_time_series_and_metadata`, which is what lets a developer materialise one asset on a laptop and have its inputs built for them, and what keeps the graph legible in the UI. What we decline is letting that edge become a runtime precondition in production. Both the principle and the rule call this out explicitly, since it's the part most likely to be "tidied up" by someone later.

## Verification

- `uv run pytest` — 395 passed, 1 skipped
- `uv run ruff check` / `ruff format --check` / `uv run --all-packages ty check` — clean
- `uv run pymarkdown scan` and `uv run mkdocs build --strict` — clean

### DIFF
--- a/docs/architecture/why-dagster-not-airflow.md
+++ b/docs/architecture/why-dagster-not-airflow.md
@@ -200,7 +200,7 @@ around the Dagster UI and would need rewriting.
 | `add_output_metadata` tables, asset catalog, lineage | every asset | Asset-event `extra` JSON (2.10+) in the events list | Partial — raw JSON, no rendered tables or history plots |
 | Asset checks — non-blocking WARN, attached to an asset, dedicated Checks view (`power_data_is_fresh`, `nwp_has_no_unexpected_nulls`) | power ingest, `ecmwf_ens` | Data-quality as ordinary tasks (`common.sql` check operators; Great Expectations / Soda / dbt-test); no first-class check primitive or Checks UI, blocking by default (as of 3.3.0) | Partial — the capability exists as tasks; the non-blocking severity and check-status surface do not |
 | `EcsRunLauncher` (laptop = subprocess, cloud = Fargate, switched by `dagster.yaml`) | control plane | ECS executor (Amazon provider, Fargate launch type) | Exists; per-*task* rather than per-run granularity |
-| Data-arrival sensors (planned, [#324](https://github.com/openclimatefix/nged-substation-forecast/issues/324)) | firing the *ingest* jobs only when there is new data to fetch, to avoid booting Fargate for a no-op | Asset-triggered DAGs, event-driven scheduling | Parity, arguably cleaner in Airflow |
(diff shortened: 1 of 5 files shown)


JackKelly:
Superseded by #454 — GitHub auto-closed this PR when its base branch `intervention-log` was deleted as part of merging #451. Same branch, same commits, rebased onto `main`.

PR #465, opened 2026-08-07
Fix every pytest warning, and make new ones fail the suite

### LINKED ISSUE #425: Fix warnings in pytest
```
15:15 $ uv run pytest
===================================== test session starts =====================================
platform linux -- Python 3.14.2, pytest-9.1.1, pluggy-1.6.0
rootdir: /home/jack/dev/python/nged-substation-forecast
configfile: pyproject.toml
plugins: hydra-core-1.3.4, anyio-4.14.2, zarr-3.3.0
collected 396 items

packages/contracts/tests/test_common.py ....                                            [  1%]
packages/contracts/tests/test_geo_schemas.py ..                                         [  1%]
packages/contracts/tests/test_hydra_schemas.py .........                                [  3%]
packages/contracts/tests/test_ml_schemas.py .......                                     [  5%]
packages/contracts/tests/test_power_forecast.py ..                                      [  6%]
packages/contracts/tests/test_power_time_series.py ......                               [  7%]
packages/contracts/tests/test_project_root.py ....                                      [  8%]
packages/contracts/tests/test_settings.py ...........                                   [ 11%]
packages/contracts/tests/test_time_series_metadata.py ......                            [ 12%]
packages/contracts/tests/test_uri.py ......                                             [ 14%]
packages/contracts/tests/test_weather_schemas.py ...                                    [ 15%]
packages/contracts/tests/test_weather_schemas_validation.py .............               [ 18%]
packages/dashboard/tests/test_forecast_chart.py ....................                    [ 23%]
packages/delta_store/tests/test_nwp.py ...                                              [ 24%]
packages/delta_store/tests/test_power_forecasts.py ....                                 [ 25%]
packages/delta_store/tests/test_precision.py ................                           [ 29%]
packages/dynamical_data/tests/test_convert_to_polars.py ............                    [ 32%]
packages/dynamical_data/tests/test_download.py ..............                           [ 35%]
packages/dynamical_data/tests/test_ecmwf_ens_cached.py .                                [ 36%]
packages/dynamical_data/tests/test_ecmwf_ens_network.py s                               [ 36%]
packages/geo/tests/test_h3.py .........                                                 [ 38%]
packages/geo/tests/test_load.py .                                                       [ 38%]
packages/ml_core/tests/test_base_forecaster.py ...                                      [ 39%]
packages/ml_core/tests/test_cross_mode_equivalence.py .                                 [ 39%]
packages/ml_core/tests/test_cv_helpers.py ............                                  [ 42%]
packages/ml_core/tests/test_feature_engineer.py ....                                    [ 43%]
packages/ml_core/tests/test_features.py .........................................       [ 54%]
packages/ml_core/tests/test_load_experiment_forecaster.py ..                            [ 54%]
packages/ml_core/tests/test_metrics.py ..............................                   [ 62%]
packages/ml_core/tests/test_mlflow_runs.py ......                                       [ 63%]
packages/ml_core/tests/test_production_helpers.py ........                              [ 65%]
packages/ml_core/tests/test_repro.py ........                                           [ 67%]
packages/nged_data/tests/test_read_nged_json.py ....                                    [ 68%]
packages/nged_data/tests/test_storage.py ..........                                     [ 71%]
packages/plotting/tests/test_ocf_theme.py ..                                            [ 71%]
packages/weather_utils/tests/test_analysis_proxy.py ............                        [ 75%]
packages/xgboost_forecaster/tests/test_forecaster.py ..............                     [ 78%]
tests/test_assets.py ............                                                       [ 81%]
tests/test_checks.py ...........                                                        [ 84%]
tests/test_cv_assets.py ......                                                          [ 85%]
tests/test_cv_power_forecasts.py ...                                                    [ 86%]
tests/test_jobs.py ......                                                               [ 88%]
tests/test_live_forecasts.py .......                                                    [ 89%]
tests/test_metrics.py ..........                                                        [ 92%]
tests/test_promoted_model.py ...                                                        [ 93%]
tests/test_register_experiment_job.py ...                                               [ 93%]
tests/test_s3_data_paths.py ...                                                         [ 94%]
tests/test_sentry.py .................                                                  [ 98%]
tests/test_trained_cv_model.py ....                                                     [100%]

====================================== warnings summary =======================================
tests/test_assets.py::test_ecmwf_ens_materialises_and_appends_nwp
tests/test_assets.py::test_ecmwf_ens_warns_on_scattered_nulls_but_still_materialises
  /home/jack/dev/python/nged-substation-forecast/.venv/lib/python3.14/site-packages/dagster/_core/storage/event_log/sql_event_log.py:3160: DeprecationWarning: datetime.datetime.utcfromtimestamp() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.fromtimestamp(timestamp, datetime.UTC).
    evaluation_event_timestamp=datetime.utcfromtimestamp(event.timestamp),

tests/test_assets.py::test_definitions_resolve
  /home/jack/dev/python/nged-substation-forecast/src/nged_substation_forecast/defs/schedules.py:39: DeprecationWarning: Parameter `partitions_def` of function `define_asset_job` is deprecated and will be removed in 2.0.0. Partitioning is inferred from the selected assets, so setting this is redundant.
    ecmwf_ens_job = define_asset_job(

tests/test_assets.py::test_definitions_resolve
  /home/jack/dev/python/nged-substation-forecast/src/nged_substation_forecast/defs/schedules.py:62: DeprecationWarning: Parameter `partitions_def` of function `define_asset_job` is deprecated and will be removed in 2.0.0. Partitioning is inferred from the selected assets, so setting this is redundant.
    live_forecasts_job = define_asset_job(

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
========================= 395 passed, 1 skipped, 4 warnings in 32.07s =========================
Exception during reset or similar
Traceback (most recent call last):
  File "/home/jack/dev/python/nged-substation-forecast/.venv/lib/python3.14/site-packages/sqlalchemy/pool/base.py", line 986, in _finalize_fairy
    fairy._reset(
    ~~~~~~~~~~~~^
        pool,
        ^^^^^
    ...<2 lines>...
        asyncio_safe=can_manipulate_connection,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/jack/dev/python/nged-substation-forecast/.venv/lib/python3.14/site-packages/sqlalchemy/pool/base.py", line 1441, in _reset
    pool._dialect.do_rollback(self)
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/home/jack/dev/python/nged-substation-forecast/.venv/lib/python3.14/site-packages/sqlalchemy/engine/default.py", line 712, in do_rollback
    dbapi_connection.rollback()
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^
sqlite3.ProgrammingError: Cannot operate on a closed database.
Exception during reset or similar
Traceback (most recent call last):
  File "/home/jack/dev/python/nged-substation-forecast/.venv/lib/python3.14/site-packages/sqlalchemy/pool/base.py", line 986, in _finalize_fairy
    fairy._reset(
    ~~~~~~~~~~~~^
        pool,
        ^^^^^
    ...<2 lines>...
        asyncio_safe=can_manipulate_connection,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/jack/dev/python/nged-substation-forecast/.venv/lib/python3.14/site-packages/sqlalchemy/pool/base.py", line 1441, in _reset
    pool._dialect.do_rollback(self)
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/home/jack/dev/python/nged-substation-forecast/.venv/lib/python3.14/site-packages/sqlalchemy/engine/default.py", line 712, in do_rollback
    dbapi_connection.rollback()
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^
sqlite3.ProgrammingError: Cannot operate on a closed database.
```

JackKelly:
Closes #425.

`uv run pytest` printed four warnings and, intermittently, two stray SQLAlchemy tracebacks after
the summary line. Three distinct causes, three treatments.

## 1. Our own bug — fixed the code

`define_asset_job` deprecates `partitions_def` ("Partitioning is inferred from the selected assets,
so setting this is redundant"), and `defs/schedules.py` passed it at both partitioned call sites
(`ecmwf_ens_job`, `live_forecasts_job`). Removed, along with the two imports that became unused.

This is a behaviour change dressed as a deprecation fix, so it is asserted rather than assumed.
`test_definitions_resolve` now checks that each **resolved** job's `partitions_def` **is** the very
object its asset declares:

```python
assert repo.get_job("ecmwf_ens_job").partitions_def is ecmwf_ens_partitions
assert repo.get_job("live_forecasts_job").partitions_def is live_forecast_partitions
```

A job silently resolving to `None`, or to a different cadence or start, now fails there rather than
at the next schedule tick. I also diffed the whole resolved repository (every job's `partitions_def`
plus every schedule's cron, timezone and target job) before and after the change — byte-for-byte
identical, including `live_forecasts_job_schedule`, which is derived via
`build_schedule_from_partitioned_job` and so depends on the inference working.

## 2. Third-party deprecation — narrowly suppressed

Dagster 1.13.16's `_core/storage/event_log/sql_event_log.py:3160` calls the
`datetime.utcfromtimestamp()` that Python 3.12 deprecated. Not fixable from this repo, so it gets a
`filterwarnings` entry pinned to **both** the exact message and the exact upstream module:

```
ignore:datetime\.datetime\.utcfromtimestamp\(\) is deprecated:DeprecationWarning:dagster\._core\.storage\.event_log\.sql_event_log
```

The module anchor is the whole point — it means the entry can only ever match that one upstream
call and can never mask the same deprecation appearing in our own code. The comment names the
package and version and says to delete it once Dagster moves to timezone-aware datetimes.

## 3. Teardown noise — root cause found and fixed

Not a warning at all: two `Exception during reset or similar` tracebacks ending in
`sqlite3.ProgrammingError: Cannot operate on a closed database`, printed *after* pytest's summary.

**Root cause, and it is ours.** `DagsterInstance.ephemeral()` builds an `InMemoryRunStorage` and an
`InMemoryEventLogStorage`. Each opens one SQLAlchemy connection against an in-memory SQLite database
and holds it for the life of the instance (`self._held_conn = self._engine.connect()`), and each
closes it only in `dispose()`. `DagsterInstance` has no finaliser, so an instance handed to the
garbage collector keeps both connections open until interpreter shutdown — where SQLAlchemy's
`_finalize_fairy` GC path can run *after* SQLite has already closed the underlying database, and
logs the traceback. Two tracebacks = two held connections = exactly one leaked instance.

All 31 `DagsterInstance.ephemeral()` call sites were in `tests/`, and each test created exactly one
instance, so they now take a `dagster_instance` fixture that enters the instance as a context
manager (`__exit__` calls `dispose()`) and tears it down when the test ends.

**Evidence.** A throwaway session-end probe counting live in-memory storages whose `_held_conn` is
still open, run over `tests/`:

| | live in-memory storages | with an open held connection |
|---|---|---|
| before | 6 | **6** |
| after | 6 | **0** |

Same instances still reachable from pytest's own caches either way; the difference is that they are
now disposed. Note the symptom itself is intermittent — it depends on finalisation order at
interpreter shutdown and did not reproduce in ~8 full runs on this machine — which is why the fix is
verified against the leak rather than against the traceback.

## `filterwarnings = ["error"]` — enabled

With the suite clean, `filterwarnings` gains a leading `error`, so a warning introduced by our own
code fails the PR that introduces it rather than accumulating in the summary. Reasoning:

- Nothing in this repo calls `warnings.warn` or `pytest.warns`, so there is no intentional warning
  for `error` to break.
- The cost is real but acceptable: a dependency upgrade that adds a new upstream warning now fails
  loudly and needs an entry here. That is the intended trade, and it matches why the ruff rule
  selection is already pinned in this file — an upgrade should change behaviour visibly, not
  silently.
- Later entries win, so `error` goes first and both ignores are applied on top of it.
- Verified on the default suite **and** on the network-gated test (`uv run pytest --run-network`,
  which `error` also covers and which is skipped by default, so it would otherwise go unchecked).

## Verification

```
uv run ruff check .              All checks passed!
uv run ruff format .             200 files left unchanged
uv run --all-packages ty check   All checks passed!
uv run pytest                    395 passed, 1 skipped in 49.80s
```

No warnings summary, and no stray tracebacks after it.

### DIFF
--- a/docs/architecture/testing.md
+++ b/docs/architecture/testing.md
@@ -67,6 +67,26 @@ row counts wrapping past 2³² rows.
   frame as well. `scripts/run_baseline_experiment.py` is the worked example. Measured with an
   `atexit` probe counting storages whose held connection is still open at exit, after one real
   `materialize`: 2 open with a bare call, 0 with the context manager, on both paths.
+- **`build_asset_context()` (and the other `build_*_context()` direct-invocation helpers) needs
+  the same treatment when called without an explicit `instance=`.** Its docstring says it
+  "Defaults to `DagsterInstance.ephemeral()`" — built internally and owned by an `ExitStack` that
+  only closes on `__exit__`, or otherwise on `__del__`. Called bare and passed straight into an
+  asset (`some_asset(build_asset_context(...))`), nothing ever calls `__exit__`, so disposal
+  depends on `__del__` running — and inside a `pytest.raises(...) as exc_info:` block, the
+  captured traceback keeps the asset's frame (and the context argument in it) referenced past the
+  end of that `with` block, so `__del__` doesn't fire until *something* triggers a GC pass, which
+  is exactly the non-deterministic timing this page keeps warning about. Enter it as a context
+  manager instead: `with build_asset_context(...) as context, pytest.raises(...) as exc_info:`.
+  `tests/test_assets.py::test_ecmwf_ens_retries_when_run_not_yet_available` is the worked example
+  — before the fix, probing immediately after that one test's teardown, with **no forced
+  `gc.collect()`**, found 2 open connections; after, 0, every time.
+- **`materialize()` and `JobDefinition.execute_in_process()` do *not* need this.** Called without
+  `instance=`, both build their default ephemeral instance behind their own internal `with
+  ephemeral_instance_if_missing(instance):`, entered and exited inside the call, before the
+  function returns — so disposal there is already deterministic regardless of how the caller uses
+  the return value. The distinguishing question for any Dagster helper that can default-construct
+  an instance is whether *the helper itself* opens and closes the `with` block, or hands you an
+  object that expects *you* to.
 
 ## Warnings are errors
 
--- a/docs/architecture/testing.md
+++ b/docs/architecture/testing.md
@@ -50,6 +50,30 @@ row counts wrapping past 2³² rows.
   (`monkeypatch.setattr(some_module, "open", fake_open)`) through the built-in fixture. For S3,
   drive the in-process `moto` server instead of mocking — `tests/test_s3_data_paths.py` is the
   canonical pattern.
+- **Take the `dagster_instance` fixture; never call `DagsterInstance.ephemeral()` in a test.** The
+  fixture lives in `tests/conftest.py`. An ephemeral instance's in-memory run storage and event-log
+  storage each hold one SQLAlchemy connection to an in-memory SQLite database open for the life of
+  the instance, and `DagsterInstance` has no finaliser — so an instance that is never `dispose()`d
+  survives to interpreter shutdown, where SQLAlchemy's connection-pool finaliser can run after
+  SQLite has closed the database and print a bare `Exception during reset or similar` traceback
+  *after* pytest's summary line, owned by no test. The fixture enters the instance as a context
+  manager, so `dispose()` runs when the test ends. Outside the test suite the equivalent is
+  `with DagsterInstance.ephemeral() as instance:`.
+
+## Warnings are errors
+
+`[tool.pytest.ini_options] filterwarnings` in `pyproject.toml` starts with `error`, so **any**
+warning fails the test that raised it. A deprecation introduced by our own code is therefore caught
+by the PR that introduces it, instead of accumulating in the warnings summary.
+
+The exceptions listed after `error` are third-party deprecations we cannot fix from this repo. Each
+one is pinned to the exact warning message *and* the exact upstream module that raises it — the
+module anchor is what stops an entry from ever masking the same deprecation appearing in our own
+code — and carries a comment naming the package, the version, and the condition for deleting it.
+
+When a dependency upgrade introduces a new upstream warning, the suite fails loudly. Add a new
+entry in that same form rather than widening an existing one; if the warning comes from our code,
+fix the code. Later entries take precedence, so `error` stays first.
 
 ## Network-gated tests
 
--- a/docs/architecture/testing.md
+++ b/docs/architecture/testing.md
@@ -59,13 +59,14 @@ row counts wrapping past 2³² rows.
   *after* pytest's summary line, owned by no test. The fixture enters the instance as a context
   manager, so `dispose()` runs when the test ends.
 - **In a script, use the context manager directly** — `with DagsterInstance.ephemeral() as
-  instance:` — rather than a bare call, so `dispose()` also runs when a step raises. A script's
-  happy path is already safe, because the local is reclaimed when the function returns; the case
-  that leaks is an *unhandled exception*, whose traceback keeps the raising frame, and therefore
-  the instance, alive until the interpreter exits.
-  `scripts/run_baseline_experiment.py` is the worked example. Measured with an `atexit` probe
-  counting storages whose held connection is still open at exit: on the error path, 2 open before
-  the context manager and 0 after.
+  instance:` — rather than a bare call. Letting the local go out of scope is *not* enough: once the
+  instance has actually run a job, Dagster's own caches retain it (a `RunDomain`, and the
+  partition-loading contexts that hold it as `dynamic_partitions_store`), so it survives to
+  interpreter shutdown with both connections open even on a completely successful run. The context
+  manager also covers the failure path, where an unhandled exception's traceback pins the raising
+  frame as well. `scripts/run_baseline_experiment.py` is the worked example. Measured with an
+  `atexit` probe counting storages whose held connection is still open at exit, after one real
+  `materialize`: 2 open with a bare call, 0 with the context manager, on both paths.
 
 ## Warnings are errors
 
--- a/docs/architecture/testing.md
+++ b/docs/architecture/testing.md
@@ -57,8 +57,15 @@ row counts wrapping past 2³² rows.
   survives to interpreter shutdown, where SQLAlchemy's connection-pool finaliser can run after
   SQLite has closed the database and print a bare `Exception during reset or similar` traceback
   *after* pytest's summary line, owned by no test. The fixture enters the instance as a context
-  manager, so `dispose()` runs when the test ends. Outside the test suite the equivalent is
-  `with DagsterInstance.ephemeral() as instance:`.
+  manager, so `dispose()` runs when the test ends.
+- **In a script, use the context manager directly** — `with DagsterInstance.ephemeral() as
+  instance:` — rather than a bare call, so `dispose()` also runs when a step raises. A script's
+  happy path is already safe, because the local is reclaimed when the function returns; the case
+  that leaks is an *unhandled exception*, whose traceback keeps the raising frame, and therefore
+  the instance, alive until the interpreter exits.
+  `scripts/run_baseline_experiment.py` is the worked example. Measured with an `atexit` probe
+  counting storages whose held connection is still open at exit: on the error path, 2 open before
+  the context manager and 0 after.
 
 ## Warnings are errors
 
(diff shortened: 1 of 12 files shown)


JackKelly:
## Review pass — what changed, and one thing deliberately left alone

An adversarial review of the first commit raised six findings. Five are addressed in 2fdeb75; the
sixth was flagged as out of scope — Jack has since asked for it, and it is now fixed (see below).

**Fixed**

1. **Both ignore entries were unanchored prefixes.** `module` is matched with `re.match`, so
   `…sql_event_log` also accepted a hypothetical `…sql_event_log_v2`. Both entries now end in `$`.
2. **The pre-existing mlflow entry did not meet the standard my own new comment block asserts** — no
   version, no delete condition, an unescaped `.` in the message, and a whole-package `mlflow.*`
   anchor. Now pinned like the Dagster one. Worth noting: the review asserted the exact raising
   module was `mlflow.utils.yaml_utils`; that is what the traceback shows, but it is wrong.
   `codecs.open()` is called from **two** mlflow modules, and anchoring on `yaml_utils` alone fails
   five tests. The entry enumerates `mlflow\.utils\.(file_utils|yaml_utils)$`.
3. **The riskiest claim was untested.** `live_forecasts_schedule` is built by
   `build_schedule_from_partitioned_job`, so its cron is *derived* from the inferred
   `partitions_def` — exactly what removing the explicit argument could have broken — and nothing
   asserted it. `test_definitions_resolve` now pins the resolved schedule's cron and timezone, and
   `live_forecasts_job` has been added to the asset-selection loop it was missing from.
4. **`is` → `==` on the new partitions assertions.** Identity is stronger than the property under
   test. `==` still catches `None` and any cadence/start difference, and is what Dagster's own
   inference uses internally, so it will not go red if Dagster ever returns an equal copy.
5. **Documented both new repo-wide rules in `docs/architecture/testing.md`** — they were previously
   discoverable only from a docstring and a TOML comment. Plus a stray blank line from the scripted
   refactor.

---

## ~~Not fixed — out of scope~~ → **Now fixed** (53d717e, diagnosis corrected in afb91df)

> ~~`scripts/run_baseline_experiment.py:124` has the same leak this PR diagnoses… I left it alone
> because CLAUDE.md says not to make out-of-scope changes without asking.~~

Jack asked for it, so it is done: `main()` now uses `with DagsterInstance.ephemeral() as instance:`,
and the five pipeline steps moved into `_run_pipeline(instance)` rather than gaining a level of
indentation.

**My original description of this leak was wrong, and measuring properly corrected it.** I first
claimed the script's happy path was safe (the local is reclaimed on return) and that only an
unhandled exception could keep the instance alive to shutdown. That measurement stubbed
`materialize` out — and stubbing `materialize` is exactly what removed the retention. With a real
materialize, Dagster keeps the instance reachable for the rest of the process (a `RunDomain`, plus
the partition-loading contexts holding it as `dynamic_partitions_store`, confirmed via
`gc.get_referrers`).

Re-measured with one real `materialize`, counting storages whose held connection is still open at
interpreter exit:

| | bare call | with context manager |
|---|---|---|
| happy path | **2 open** | 0 open |
| error path | **2 open** | 0 open |

So the context manager is not error-path insurance — it is the only thing that closes these
connections at all. `docs/architecture/testing.md` and the code comment were corrected in afb91df;
the code fix itself was right from the start.

Observable behaviour is unchanged: happy-path stdout is byte-identical, and on failure the exception
still propagates with exit code 1, skipping `_report_metrics()` and the closing "Done." line. The
other two Python scripts create no Dagster instance and no DB connection, so this was the only
remaining site.

Issue #469, opened 2026-08-07
Delete the local model cache in load_from_mlflow


JackKelly:
## The mechanism

`BaseForecaster.load_from_mlflow` keeps a local model cache under
`settings.model_cache_base_path`, keyed by MLflow run ID. On a cache hit it loads straight from
disk and never contacts MLflow.

The cache key is a run ID, which is immutable — but **a CV fold run's artifacts are not**.
`get_or_create_fold_run` reuses one run per `(experiment, fold)` across materialisations, so
re-training a fold replaces its artifacts under an unchanged ID. That is a silent-wrong-model
bug: a re-trained fold would be scored with the model it had just replaced.

PR #464 closes it by making `save_to_mlflow` invalidate the run's cache entry on write, which
means the write path now takes the read path's storage location as a required argument.

## Why delete it instead

The cache has exactly **one** caller: `cv_assets.py:500`, in `cv_power_forecasts`, once per
materialisation and outside the chunk loop. The live service does not use it — `live_forecasts`
loads from a plain disk directory via `load_forecaster_from_dir`
(`defs/production_assets.py:238`), and `Settings.model_cache_base_path`'s own description says
so: *"production inference does not use it for v0.1 (the model is baked directly into the
container image instead)"*.

So the cache's entire present value is avoiding one model download per fold, on a laptop, from a
local `file://` or SQLite store. In exchange it produced the staleness bug above, and now
requires a write-path/read-path coupling to keep it honest. It has not earned its operating
cost.

Deleting it also closes a second known defect for free: an interrupted download currently
poisons the cache permanently, because a partial directory counts as a cache hit. With no cache,
`load_from_mlflow` collapses to "download to a temp dir, load" and that failure mode cannot
exist.

## Proposed work

- `load_from_mlflow` downloads to a temporary directory and loads; no cache, no cache key.
- Remove the `cache_base_path` parameter from `load_from_mlflow` **and** the `cache_base_path`
  keyword and `_delete_cache_entry` helper that PR #464 adds to `save_to_mlflow`.
- Remove `Settings.model_cache_base_path` and any `.env` / docs references, unless something
  else uses it (check).
- Update `docs/architecture/ml-orchestration.md` (the model-artifacts section) and
  `docs/architecture/production-deployment.md` — the latter currently names this cache as the
  *future* production-resilience mechanism, which is the claim #REPLACE_D exists to revisit.

## Note

[...]

Issue #197, opened 2026-06-30
Make trained_cv_model fold-run param logging re-run-safe (MLflow param immutability)


JackKelly:
## Problem

Re-materialising `trained_cv_model` into an **existing** MLflow fold run fails at the final `mlflow.log_params(...)` with:

```
MlflowException: Changing param values is not allowed. Params were already logged=
'[{key: n_eligible_time_series, old_value: 0, new_value: 28}]' for run ID=...
```

`get_or_create_fold_run` reuses the fold run by tag, and MLflow params are immutable. So when a fold is re-run with different inputs (e.g. a run created during the earlier empty-eligible no-op era logged `n_eligible_time_series=0`, then a re-run after the data/val_end fix logs `28`), the param write is rejected.

`save_to_mlflow` runs **before** the param logging, so the model artifacts *are* saved — but the asset still errors at the end, so the Dagster run is marked failed and downstream metrics aren't logged.

## Repro

1. Materialise `trained_cv_model` for a fold when its eligible set is empty (logs `n_eligible_time_series=0`).
2. Fix the data so the fold has eligible series; re-materialise the same `(experiment, fold)` partition.
3. The run trains + saves the model, then raises on `mlflow.log_params`.

## Fix options

- Log mutable counters (`n_eligible_time_series`, `n_trained_time_series`) via `mlflow.set_tags`/metrics rather than params, or
- Make param logging idempotent (skip/overwrite when already present), or
- Start a fresh fold run per materialisation instead of reusing by tag.

## Notes

- Code: `src/nged_substation_forecast/defs/cv_assets.py` (`trained_cv_model`).
- Related gotcha: `load_from_mlflow` caches by the immutable run_id, so re-training a reused fold run also leaves a stale local model cache (must be cleared to pick up the new artifacts).
- Discovered while verifying #91 (PR #195).

Design: [docs/roadmap/live-service.md → "Related GitHub issues"](https://openclimatefix.github.io/nged-substation-forecast/roadmap/live-service/#related-github-issues)

JackKelly:
A similar issue (but not the same issue) was fixed in PR #202. This issue is about trained_cv_model's mlflow.log_params (for counters like n_eligible_time_series, n_trained_time_series), whereas PR #202 only fixes cv_power_forecasts's log_params for val_start/val_end.

PR #482, opened 2026-08-07
Upload the trained model to MLflow as a single replaceable archive artifact

### LINKED ISSUE #470: Upload the trained model to MLflow as a single replaceable archive artifact
## The problem

`BaseForecaster.save_to_mlflow` uploads the model with `mlflow.log_artifacts`, which **merges**
into the run's artifact directory rather than replacing it. Verified empirically against MLflow
3.15.1: uploading `{a.ubj, meta.json}` over an existing `{a.ubj, b.ubj, meta.json}` leaves
`b.ubj` in place.

Because a CV fold run is reused across materialisations, re-training a fold on a **smaller**
population leaves the dropped series' model files behind in the run. MLflow exposes no public
artifact-delete API, so they cannot be cleaned up in place.

PR #464 defuses the immediate danger by having `XGBoostForecaster.load` take its population from
`meta.json` instead of globbing `*.ubj` — the right change regardless, since a model's population
is the model's own record and not a directory listing. But the stale files still accumulate
inside the run, and they are still downloaded and baked into the container image by
`fetch_model_artifacts`.

## Proposed work

Upload the model as a **single archive artifact** (one file, e.g. a `.tar.gz`) rather than a
directory. A same-name single artifact overwrites cleanly — verified empirically — so a re-upload
replaces rather than merges, and the stale-file problem cannot arise.

- `save_to_mlflow` archives the temp directory and logs the one file.
- `load_from_mlflow` and `_production_helpers.fetch_model_artifacts` gain an unpack step.
- Subclass `save`/`load` stay directory-based and MLflow-free, so `load_forecaster_from_dir` and
  the Docker bake are untouched.

The only real cost is losing per-file browsing of model artifacts in the MLflow UI, which is
negligible for a directory of `.ubj` blobs.

Sequencing: this touches the same functions as #469 (delete the model cache). Doing that
one first makes this one smaller, since `load_from_mlflow` will already have collapsed to
download-and-load.

## Provenance

Architecture review of [PR #464](https://github.com/openclimatefix/nged-substation-forecast/pull/464)
(issue #197), which reports the accumulation as a known-remaining item and correctly declines to
fix it in place.

JackKelly:
Closes #470.

## The problem

`BaseForecaster.save_to_mlflow` uploaded the model with `mlflow.log_artifacts`, which **merges**
into a run's artifact directory rather than replacing it. Because a CV fold run is reused across
re-materialisations, re-training a fold on a **smaller** population left the dropped series' model
files behind in the run — and MLflow exposes no public artifact-delete API, so they could never be
cleaned up. They were then downloaded on every subsequent load and baked into the container image
by `fetch_model_artifacts`.

PR #464 (issue #197) defused the *danger* by having `XGBoostForecaster.load` read its population
from `meta.json` rather than globbing `*.ubj`, so a stale file can no longer be silently served.
That fix stands unchanged; this PR stops the accumulation at the source.

## The change

`save_to_mlflow` packs the model directory into a single `model.tar.gz` and logs **that one file**
to the run's artifact root. A same-name single artifact overwrites cleanly, so a re-upload
replaces rather than merges.

- `load_from_mlflow` and `ml_core._production_helpers.fetch_model_artifacts` share one
  `_download_and_unpack_model` helper, so the archive layout is defined in exactly one place.
- Subclass `save`/`load` stay directory-based and MLflow-free — `load_forecaster_from_dir` and
  the Docker bake path are untouched by this change.
- `_production_helpers` no longer imports `mlflow` at all.

### Verified empirically (MLflow 3.15.1)

Re-confirmed the issue's premise directly against a file-based tracking store:

| upload | before | after re-upload of a *smaller* model |
|---|---|---|
| `log_artifacts` (directory) | `{a.ubj, b.ubj, meta.json}` | `{a.ubj, b.ubj, meta.json}` — `b.ubj` **survives** |
| `log_artifact` (one same-name file) | `model.tar.gz` (20 bytes) | `model.tar.gz` — contents **replaced** |

## Cost

Per-file browsing of model artifacts in the MLflow UI is lost; the run now shows one
`model.tar.gz` instead of a browsable `model/` directory. Negligible for a directory of opaque
`.ubj` blobs, and noted as such in the docs.

## Tests

`packages/ml_core/tests/test_base_forecaster.py` — the fake forecaster now writes one file per
`time_series_id` (mirroring `XGBoostForecaster`'s one-`.ubj`-per-series layout), which is what
makes a shrinking population observable. Added:

- `test_the_model_is_stored_as_a_single_archive_artifact` — the run holds exactly one model file.
- `test_re_saving_a_smaller_model_leaves_no_trace_of_the_dropped_series` — **the regression this
  issue exists to prevent**. Re-saves a smaller population into the same run and asserts the
  dropped series is absent from the run's artifact listing, from `load_from_mlflow`'s population,
  and from what `fetch_model_artifacts` lands on disk. Confirmed to fail against the pre-fix
  `log_artifacts` implementation.
- `test_fetch_model_artifacts_unpacks_the_archive_into_dest` — the production download path lands
  a plain model directory, not an archive file.

The `fetch_model_artifacts` tests live in this file rather than `test_production_helpers.py`
because they exercise the same archive plumbing and share the fake forecaster. The existing
end-to-end coverage of the download path with a real `XGBoostForecaster`
(`tests/test_promoted_model.py`) is unchanged and still passes.

Existing tests were updated only where their docstrings described the now-removed merge hazard;
none were relaxed. `ruff check`, `ruff format`, `ty check`, `pytest` (443 passed, 1 skipped),
`pymarkdown` and `mkdocs build --strict` all green.

## Docs

`docs/architecture/ml-orchestration.md` gains a "One archive file, not a directory of files"
subsection with the rationale and the empirical result. The section heading changed, so its
anchor did too — all five inbound links (two docstrings, three docs pages) were retargeted.

**Adjacent one-line fix:** the same section had a pre-existing instance of the wrapped-`#` heading
gotcha documented in CLAUDE.md — `(issue` / `#469)` split across a line break rendered `#469)` as
an `<h1>` and broke the paragraph in two. Rewrapped so the `#` is no longer at line start;
verified by reading the generated HTML under `site/`.

### DIFF
--- a/docs/architecture/ml-orchestration.md
+++ b/docs/architecture/ml-orchestration.md
@@ -41,15 +41,48 @@ The complementary decision — the resolved config is stamped onto the MLflow ex
 registration and read back by the assets, never re-read from YAML — is explained in
 [Running an experiment end-to-end](../ml_experimentation/dagster-workflow.md).
 
-## Model artifacts: MLflow artifact store, no local cache
+## Model artifacts: one replaceable archive, no local cache
 
 Trained models live in MLflow's artifact store, wrapped by two concrete `BaseForecaster`
 methods (`save_to_mlflow` / `load_from_mlflow`) that delegate to each subclass's own disk
 `save`/`load` — subclasses stay MLflow-free. `load_from_mlflow` downloads straight into a
 temporary directory and loads from there; there is no local-disk cache.
 
-An earlier version cached downloads on disk keyed by MLflow run ID. That was removed (issue
-#469) because it was actively harmful rather than merely unused: a CV fold run is **reused**
+### One archive file, not a directory of files
+
+`save_to_mlflow` packs the model directory into a single `model.tar.gz` and logs that one file
+to the run's artifact root. `load_from_mlflow` and the production download path
+(`ml_core._production_helpers.fetch_model_artifacts`) each unpack it into a temporary directory.
+Subclass `save`/`load` never see the archive: they stay directory-based and MLflow-free, so the
+Docker bake path (`load_forecaster_from_dir`) is unaffected.
+
+The archive exists because **MLflow's directory upload merges rather than replaces**, and MLflow
+exposes no public artifact-delete API. Verified empirically against MLflow 3.15.1: re-uploading
+`{a.ubj, meta.json}` with `log_artifacts` over an existing `{a.ubj, b.ubj, meta.json}` leaves
+`b.ubj` in place, while re-logging a same-name single artifact overwrites it outright. A fold run
+is **reused** across re-materialisations of its partition (see "Cross-process run resolution"
+above), so with a directory upload, re-training a fold on a *smaller* population would strand the
+dropped series' model files in the run permanently — downloaded on every subsequent load and
+baked into the container image. One replaceable file makes that accumulation impossible at the
+source (issue #470).
+
+Two properties are worth keeping in mind when reading the code:
+
+- The run holding exactly one model file is the load-bearing invariant, not an implementation
+  detail. Logging any *second* per-model file alongside the archive would reopen the merge
+  problem for that file.
+
+- A model's population is still read from its own `meta.json`, never from a directory listing —
+  the archive removes the accumulation, but the population contract stands on its own (the
+  train==predict invariant; see `BaseForecaster.trained_time_series_ids`).
+
+The cost is losing per-file browsing of model artifacts in the MLflow UI, which is negligible for
+a directory of opaque model blobs.
+
+### Why there is no local cache
+
+An earlier version cached downloads on disk keyed by MLflow run ID. That was removed
+(issue #469) because it was actively harmful rather than merely unused: a CV fold run is **reused**
 across re-materialisations of its partition (see "Cross-process run resolution" above), so the
 same run ID can legitimately hold a different model after re-training, which made the cache key
 non-unique for its contents. Keeping the cache honest under that reuse required write-side
--- a/docs/architecture/ml-orchestration.md
+++ b/docs/architecture/ml-orchestration.md
@@ -76,8 +76,18 @@ Two properties are worth keeping in mind when reading the code:
   the archive removes the accumulation, but the population contract stands on its own (the
   train==predict invariant; see `BaseForecaster.trained_time_series_ids`).
 
-The cost is losing per-file browsing of model artifacts in the MLflow UI, which is negligible for
-a directory of opaque model blobs.
+Two costs come with this. The first is losing per-file browsing of model artifacts in the MLflow
+UI, which is negligible for a directory of opaque model blobs. The second is that both sides now
+hold the model twice at once in their temporary directory — the model directory plus the archive
+when saving, the archive plus its unpacked contents when loading — so a machine running these
+assets needs roughly twice the model's size in free temp space. At V2 scale (~2,500 series) that
+is a few GB, which matters mainly if `TMPDIR` is a memory-backed filesystem inside a container.
+
+The archive is written at gzip level 1 rather than `tarfile`'s default of 9. Measured on 40 real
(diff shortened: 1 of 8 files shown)


JackKelly:
## Migration note

**Every MLflow fold run written before this PR must be re-trained.** `load_from_mlflow` and
`fetch_model_artifacts` now look for `model.tar.gz`, which no existing run has, so
re-materialising `cv_power_forecasts`/`metrics` on an already-trained fold, or promoting an
existing champion, will fail until `trained_cv_model` is re-materialised for that fold. The error
says so explicitly rather than surfacing MLflow's bare "path not found".

## Review round

An independent adversarial review raised the following. Fixed in 3f93152:

1. **gzip level (the one substantive finding).** `tarfile.open(..., "w:gz")` defaults to level 9,
   the slowest. Re-measured independently on 40 real boosters (500 estimators, depth 6, 24
   features; 77 MB of `.ubj`):

   | level | time | size | ratio | extrapolated to ~2,500 series |
   |---|---|---|---|---|
   | 1 | 0.68 s | 28.4 MB | 2.73x | 0.7 min |
   | 4 | 0.82 s | 24.7 MB | 3.13x | 0.8 min |
   | 6 | 2.27 s | 23.0 MB | 3.37x | 2.4 min |
   | 9 (default) | 14.85 s | 22.3 MB | 3.47x | **15.5 min** |

   The default would spend ~15 minutes of CPU per fold at V2 scale to save a fifth of the bytes,
   on the training asset's critical path. Now pinned to level 1 via `_ARCHIVE_COMPRESSLEVEL`,
   with the measurement in the constant's docstring and in the docs.

2. **Pre-existing runs fail opaquely.** `_download_and_unpack_model` now re-raises with the cause
   and the remedy; new test `test_loading_a_run_with_no_archive_says_what_to_do_about_it`.

3. **Doubled temp-disk requirement** (model dir + archive when saving, archive + unpacked
   contents when loading) — now stated in the docs' cost paragraph, which previously mentioned
   only the MLflow UI.

4. **Docstring nits** — `BaseForecaster._MLFLOW_MODEL_ARTIFACT` is a module-level constant, not a
   class attribute; "two consumers"/"third consumer" read as a contradiction. Both fixed.

5. **The whole-artifact-listing assertion is stricter than the invariant it names.** Kept as-is
   deliberately, with the reason now written into the test docstring: nothing else logs artifacts
   to a fold run today, and a second artifact showing up here deserves a deliberate look rather
   than a silent pass, since any *per-model* file logged alongside the archive would reopen the
   merge problem for that file.

### Raised and not changed

- **`model_dir.mkdir()` before `save` is redundant**, since `XGBoostForecaster.save` creates its
  own directory. Kept: it preserves the previous contract exactly — `save_to_mlflow` has always
  handed `save` a directory that already exists — rather than making every future subclass
  responsible for creating it.

- **A bare `"model"` literal for the temp sub-directory** while the unpack directory had a
  module constant. Resolved the asymmetry the other way: the unpack name is now a local literal
  with a comment on the only thing that constrains it (it must not collide with the downloaded
  archive's own name), since neither name is part of any contract.

- **`### Why there is no local cache` is written as history**, which is the changelog prose
  CLAUDE.md warns against. Correctly identified as pre-existing — this PR only moved and
  re-headed the passage. Left alone as out of scope; worth a separate look, since the rationale
  it records (issue #469) may be worth keeping in some form.

### Confirmed clean by the reviewer

`tar.extractall(..., filter="data")` is safe (no CVE-2007-4559 exposure); nested sub-directories
round-trip; the regression test genuinely fails against the pre-fix implementation (independently
verified by monkeypatching the old `log_artifacts` path back in); no existing test was relaxed;
no other code in the repo reads or writes the run artifact layout; all five inbound anchor links
retarget correctly.

`main` has also been merged into this branch (no conflicts). Full suite: 490 passed, 1 skipped.

PR #517, opened 2026-08-10
Note the sticky closing-issue links in the PR-workflow skill

### LINKED ISSUE #512: Reject unknown keys in config_overrides (extra="forbid" on BaseForecasterConfig)
> 🤖 **This issue was written by [Claude Code](https://claude.com/claude-code)**, acting on Jack's behalf.

`BaseForecasterConfig` is a plain pydantic `BaseModel`, so it inherits pydantic's default
`extra="ignore"`. A `config_overrides` key that does not name a real field is silently dropped:

```json
{"n_estimtors": 5000}
```

registers happily, trains with `n_estimators=500` from the base YAML, and produces a run that
looks like a valid experiment.

## Why this matters more than a typo usually would

Both planned search mechanisms are **programmatic and unattended**, so nobody reads the resolved
config before the runs go out:

- The [LLM-driven auto-research agent](https://openclimatefix.github.io/nged-substation-forecast/roadmap/#v06-nged-consume-the-forecast) registers experiments, materialises them and reads the leaderboard with no human in the loop.
- The [training-history variant grid](https://openclimatefix.github.io/nged-substation-forecast/roadmap/training-history/) sweeps five dimensions at once.

One misspelled key in either produces a grid of *identical* runs that all score plausibly and all
land on the leaderboard. That is a [principle 8](https://openclimatefix.github.io/nged-substation-forecast/design-philosophy/design-principles/#8-every-experiment-is-scored-identically)
problem, not just an ergonomics one: the scores are real, but the experiments are not what they
claim to be, and nothing distinguishes the bad grid from a genuine null result.

It is also a [naming-poka-yoke](https://openclimatefix.github.io/nged-substation-forecast/design-philosophy/design-principles/#naming-poka-yoke) case — make the wrong usage fail to parse rather than merely be discouraged.

## Proposed change

Add `model_config = ConfigDict(extra="forbid")` to `BaseForecasterConfig`
(`packages/ml_core/src/ml_core/base_forecaster.py`), so an unknown key raises `ValidationError` at
registration time, before any fold runs.

Check before landing: `_resolve_forecaster_config` pops both `_target_` keys before construction,
so those must not trip the new strictness; and `conf/model/xgboost.yaml`'s `model_params` must
contain no key that is not a declared field.

## Out of scope

Not part of [#228](https://github.com/openclimatefix/nged-substation-forecast/issues/228) — the
behaviour is identical before and after the Hydra removal (Hydra's `instantiate` also just called
the constructor). Filed separately so the removal stays a pure swap.

JackKelly:
> 🤖 **This PR description was written by [Claude Code](https://claude.com/claude-code)**, acting on Jack's behalf.

Adds a "check what the merge will close" step to the `github-issue-pr-workflow` skill.

## Why

Issue [#512](https://github.com/openclimatefix/nged-substation-forecast/issues/512) was closed by
the merge of [#514](https://github.com/openclimatefix/nged-substation-forecast/pull/514) — a PR
that had deliberately left it alone, and whose body said so ("Two things noticed in passing, filed
rather than fixed"). It has since been reopened, with its project Status put back to Todo.

The cause is that a closing reference is sticky. GitHub registers the link when the text
containing the keyword is first saved, and does **not** remove it when the text is edited away.
An early draft of #514's body must have said "closes #512"; by merge time no trace of that
remained in either the body or the branch's five commit messages, but the link was still live:

```console
$ gh pr view 514 --json closingIssuesReferences --jq '.closingIssuesReferences[].number'
228
512
```

So the check has to be that field. Reading the body is not evidence, and neither is grepping the
commits.

## What the note says

The command to run before `gh pr merge`, why the current body cannot be trusted, and how to
recover — because reopening is not the whole of it. The board automation moves a closed issue to
Done, and a reopen lands it on **In Progress**, not back on Todo, so the project field needs
setting explicitly.

### DIFF
--- a/.claude/skills/github-issue-pr-workflow/SKILL.md
+++ b/.claude/skills/github-issue-pr-workflow/SKILL.md
@@ -50,6 +50,23 @@ Never squash-merge. Jack wants the full commit history preserved in `main`, so u
 (`gh pr merge --merge`) or rebase (`gh pr merge --rebase`), not `gh pr merge --squash`. Under the
 `implement-issue` routine you stop and wait for Jack's review rather than merging at all.
 
+**Check what the merge will close, before you merge:**
+
+```bash
+gh pr view <N> --json closingIssuesReferences --jq '.closingIssuesReferences[].number'
+```
+
+Every number listed is closed the moment the PR merges. The list is *sticky*: a closing keyword
+in an early draft of the PR body, or in any commit message on the branch, registers the link
+permanently, and later editing that text away does not remove it. So a PR whose body now says
+"filed rather than fixed, see #512" can still be holding a closing link to #512 from a draft —
+reading the current body is not enough, and neither is grepping the commits.
+
+If the list contains an issue you did not mean to close, either sort it out before merging or
+watch for it afterwards: `gh issue reopen <N>`, then put its project Status back (the board
+automation moves a closed issue to Done, and reopening it lands on In Progress, not Todo — see
+the `github-graphql` skill for `gh project item-edit`).
+
 ## GraphQL calls
 
 Attaching and reordering sub-issues, setting an issue's Type, and setting a project field all need


JackKelly:
Superseded: this PR's first body used a closing keyword next to the issue reference and so registered a closing link that a body edit would not clear. Reopening as a fresh PR from the same branch.

Issue #446, opened 2026-08-05
Make live_forecasts degrade rather than raise when NWP is absent


JackKelly:
Make `live_forecasts` **degrade rather than raise** when NWP is absent or too old to cover the forecast horizon.

**Today's behaviour.** A weeks-old NWP run no longer covers the forecast window, so its rows join to a null `ensemble_member`, are filtered out, and the asset raises on an empty frame (`production_assets.py`). This is rung 2 of the [degradation ladder](https://openclimatefix.github.io/nged-substation-forecast/design-philosophy/inherent-stability/#the-degradation-ladder), and NGED get nothing at all. `select_nwp_init_time` raises in the same class of situation (`_production_helpers.py`) — narrower in practice, since it fires only when the NWP table holds no run at or before the cutoff, but it bites `replay` backfills older than the retention window.

**Keep the `trained_ids` raise.** An empty promoted model is a promotion bug, not a data outage, and it should stay loud. The codebase already distinguishes the two cases; it just does not do so consistently.

**Ordering matters, in both directions.**

- The **missed-NWP-runs asset check** ([#424](https://github.com/openclimatefix/nged-substation-forecast/issues/424)) must land *first*. Fixing this raise without it would convert a loud failure into a silent one, which is strictly worse: today's second divergence is that NWP between 0 and ~15 days stale already degrades with no warning anywhere.
- It must land *after* the weather-blind training work. Before the model is trained for outages, a no-NWP forecast rides untested default directions, so degrading early would emit unvalidated output. Ship once scenario scoring shows rung-2 skill beats `manual_heuristic`, or together with that training work.

Until then, loud failure plus the asset check is strictly better than today's silence — and there is no external consumer before ~v0.6 anyway, so the urgency is lower than it looks.

**The real work is deciding *what* a no-NWP forecast contains**, not removing the raises.

JackKelly:
> 🤖 **This comment was written by [Claude Code](https://claude.com/claude-code)**, acting on Jack's behalf.

One of the two blockers named above has landed: the missed-NWP-runs asset check (#424) is now closed. The remaining blocker is the weather-blind training work (#445) — that gate still stands, for the reason given above: degrading early would emit unvalidated output before scenario scoring shows rung-2 skill beats `manual_heuristic`.

[...]

PR #531, opened 2026-08-11
Refuse to promote a model whose features this code cannot parse

### LINKED ISSUE #486: promoted_model should reject a model whose selected_features the current code cannot parse
## The failure this prevents

PR #463 renamed the time feature `local_utc_offset` → `local_utc_offset_minutes`. Every reference
in the repo moved with it, and the PR verified that with a full-repo grep. But the **promoted model**
in `data/production_model` still carried the old name, in both `meta.json`'s `selected_features` and
the boosters' own `feature_names`.

The result: every `live_forecasts` tick failed with

```
ValueError: Unrecognised feature name: local_utc_offset
```

raised from `ParsedFeatures.from_strings()` (`packages/ml_core/src/ml_core/features/_parsed_features.py`),
reached via `live_forecasts` → `forecaster.feature_engineer.engineer(selected_features=...)` at
`src/nged_substation_forecast/defs/production_assets.py:269`.

This was only noticed because someone happened to read a failed run's stack trace. Nothing detected
it at the point the breaking change landed, and nothing would have detected it until the next tick
failed.

## Why the existing safeguards did not catch it

- **A full-repo grep cannot see the artefact.** `data/` is gitignored (`.gitignore:29`), so
  `data/production_model/meta.json` is structurally outside the search space of the check PR #463
  used. The one live consumer of the old feature vocabulary is the one place the grep could not look.
- **`_reject_changed_identity` guards the wrong direction.** It stops a *re-registration* of an
  experiment whose config changed. It says nothing about an already-promoted model whose config the
  current code can no longer parse.
- **The "no retrain needed" reasoning was sound but incompletely applied.** PR #463 skipped a retrain
  because only a dummy model had been trained and the leaderboard was not populated — true for
  leaderboard comparability, but that dummy model was also the one being *served* every 6 hours.
  Nobody joined those two facts up.

## Proposed fix

In `promoted_model` (`src/nged_substation_forecast/defs/production_assets.py`), after
`fetch_model_artifacts` and the `meta.json` read, validate the model's feature vocabulary against
the running code before accepting it:

```python
ParsedFeatures.from_strings(set(model_params["selected_features"]))
```

and fail the materialisation with an actionable message if it raises — naming the offending feature
and telling the operator to retrain rather than promote.

Open design questions for whoever picks this up:

- **Where to validate.** Validating after the download is simplest, but leaves a broken model
  written to `production_model_path`. Validating before the atomic swap would be better — check the
  freshly-downloaded `meta.json` in the staging directory and abort the swap, so a rejected
  promotion leaves the previous working model in place. Worth checking what
  `ml_core._production_helpers.fetch_model_artifacts` makes easy here.
- **Booster feature names too?** `meta.json` and the boosters can in principle disagree. Comparing
  `Booster.feature_names` against `selected_features` would catch that, at the cost of loading the
  boosters at promotion time.

## Why raising here is correct

This does not conflict with the inherent-stability principle that production degrades rather than
raises. That rule is about the *outside world* misbehaving — absent or stale inputs. A promoted
model whose feature vocabulary the serving code cannot parse is a **contract violation of our own
making**, which the principle explicitly reserves raising for. And promotion is a manually-triggered
R&D-side action with a human watching, not an unattended production tick.

The guard turns "live forecasting is silently broken until someone reads a stack trace" into
"promotion refuses a model this code cannot serve", at the moment an operator is present to act on it.

JackKelly:
🤖 Written by [Claude Code](https://claude.com/claude-code).

Closes #486.

A saved model names its features as strings, so renaming a feature in code leaves every model trained before the rename unservable — and nothing notices, because `data/` is gitignored and so the full-repo grep a rename PR relies on structurally cannot see the promoted artefact.

`_check_selected_features_are_parseable` runs `ParsedFeatures.from_strings` over a model's `selected_features` and re-raises with the offending feature name and the remedy. Two call sites:

- **`fetch_model_artifacts`** reads the staged `meta.json` and checks it **before the atomic swap**, so a refused promotion leaves the previous champion in place and still serving.
- **`load_forecaster_from_dir`** checks the reconstructed config, so a live tick fails at model load rather than partway through feature engineering.

## Two corrections to the issue, both verified

**This guard would not have caught the incident #486 opens with.** PR #463's rename merged 2026-08-07 13:22 UTC; `data/production_model/promotion.json` shows the model currently on disk was promoted at 17:35 UTC and carries the new name. The breakage was *promote-then-change-the-code*, and a promotion-time guard runs before the breaking change. What it catches is the mirror image — promoting a model whose vocabulary the current code cannot parse. That is a genuine hazard, but a different one, and it is preventive rather than remedial: every run trained from now on is stored as a promotable archive regardless of its feature vocabulary, so the next rename recreates the trap with no gate in front of it.

**Nothing in the MLflow store today would trip it.** `list_promotable_runs()` returns three candidates and two do carry the stale `local_utc_offset` vocabulary — but both predate the single-archive change, so they hold the old `model/` directory layout and `_download_and_unpack_model` already refuses them. The guard fires on zero of today's candidates.

## Departures from the issue's proposed fix

- **The check reads `meta.json`, it does not load the model.** Routing `fetch_model_artifacts` through `load_forecaster_from_dir` would break two existing tests (`_FakeForecaster` writes `model_class: "fake"`, which `import_class` rejects), would force a model library into `ml_core`'s deliberately model-library-free test suite, and would pull every booster into memory — several GB at V2's ~2,500 series — to answer a question one JSON field answers.
- **The booster `feature_names` comparison (the issue's second open question) is rejected.** Verified against xgboost 3.4.0: `QuantileDMatrix(polars_df, …)` sets `feature_names` from the frame's columns, they survive a `.ubj` round-trip, and `Booster.predict` raises `ValueError: feature_names mismatch: …`. Since `_feature_cols` is `sorted(selected_features)` and `_prepare_features` selects exactly those columns, a `meta.json`-vs-booster disagreement is already loud and names the offending column. It is not the silent class of failure this issue exists to close, and catching it needs either a hook on the `BaseForecaster` ABC or model-specific code in a model-agnostic helper.
- **An absent `selected_features` passes rather than raising.** `BaseForecaster.save`'s contract mandates only `model_class`, so a saved record naming no features is missing input, not malformed input — rule 2 of [inherent stability](https://openclimatefix.github.io/nged-substation-forecast/design-philosophy/inherent-stability/#the-rules). Pinned by a test so a later tightening into fail-closed has to be deliberate.

## One non-obvious constraint, now recorded in the code

The error message must not contain the substring `mlflow` in any casing. `scripts/build_and_verify_image.sh` hard-fails the image build on `grep -qi "mlflow"` over the container log as its one automated hermeticity gate, and this message can reach that log via `live_forecasts`. Verified the whole traceback is clean.

## Why raising is correct here

Rule 1 reserves raising for "states that are our own bug — an empty promoted model, a contract violation". A promoted model whose vocabulary the serving code cannot parse is the same class as the empty-population raise fifteen lines away in `live_forecasts`. #446 makes the same argument in the opposite direction for NWP staleness. No asset check is added, so rules 6 and 7 are untouched: nothing warns, and no warning path gains a way to raise. Serves H1/T1.1 by removing an `our-bug` intervention-log category before it can happen.

## Tests

Three new behaviour tests, each verified to fail with the guard disabled and pass with it:

- `test_load_forecaster_from_dir_rejects_an_unparseable_feature` — pins `local_utc_offset`, the real retired name, as a regression.
- `test_fetch_model_artifacts_keeps_the_previous_model_when_the_new_one_is_unservable` — the outgoing champion and its `promotion.json` survive a refused promotion.
- `test_promoted_model_refuses_a_model_with_an_unparseable_feature` — asset-level, real file-store MLflow + Dagster. Note `materialize` defaults to `raise_on_error=True`, so this is written as `pytest.raises` rather than asserting on `result.success`.

Plus two contract pins that pass on `main` too and say so in their docstrings: a negative control covering one feature of each kind the parser understands (so a guard that rejected everything could not satisfy the test above), and the missing-vs-malformed asymmetry.

`_FakeForecaster` now writes `model_params` into its `meta.json` and reads it back on `load`, matching what every real `BaseForecaster.save` writes.

## Docs

`docs/live_service/operations.md` Steps 1–3, the failure-modes row in `docs/design-philosophy/inherent-stability.md`, and the `scripts/build_and_verify_image.sh` header — the image build is the first moment new code and a model promoted against older code meet, so its "confirm by eye" section now says what a feature-name traceback there means.

## Verification

`ruff check`, `ruff format`, `ty check`, `pytest` (588 passed, 1 skipped), `pymarkdown scan`, `mkdocs build --strict` — all green.

The plan this was built from is in `plans/`, including what the adversarial review of the plan changed and what it found that was rejected.


JackKelly:
🤖 Written by [Claude Code](https://claude.com/claude-code).

An independent adversarial review of the diff — including the mutation testing asked for — found a real hole in the first version, plus five surviving mutants. Fixed in fcd9e2dc.

**The hole.** `BaseForecasterConfig.selected_features` is a *required* field, so a saved record without it loads nowhere. The first version treated an absent `selected_features` as "missing rather than malformed input" and waved it through — which meant promotion would `rmtree` a working champion and install a model that dies at its first `load`. Exactly the outcome the pre-swap ordering exists to prevent. The rule-2 analogy in the PR body was wrong: the field is not optional at the boundary that consumes it. It is now refused, and so are a non-dict `model_params` and non-string feature entries — those used to escape as `AttributeError`/`TypeError` past the `except ValueError`, and a bare `"temperature_2m"` was set-ified into its characters and reported them as the unrecognised features.

**Other findings applied:**

- **Every unparseable feature is named now, sorted.** Reporting only the first sent an operator round the re-train loop once per retired feature — and which came first varied per process, because the names live in a `set`.
- **The load path checks before `load`, not after.** The PR argued at length that promotion must not load the model because of the V2-scale booster cost; the same argument applied verbatim to the serving path, where `meta` was already in hand two lines earlier.
- **The promotion call site passes a bare run id.** It passed `f"MLflow run {run_id}"`, which falsified the comment sitting above the raise claiming no message may contain that word. Harmless in practice (`fetch_model_artifacts` never runs in the container), but a comment that is false where it sits.

**Mutation results.** Eight mutants, all killed, each verified by editing the source and running the suite: drop the check in `load_forecaster_from_dir`; drop it in `fetch_model_artifacts`; move it after the swap; check only the first feature; report only the first offender; drop the source from the message; put the tracker's name in the message; accept an absent feature list; accept non-string entries.

Two mutants survive and are **equivalent**, not coverage holes — recorded so the next reviewer does not re-derive them. Writing `promotion.json` before the check rather than after is unobservable: it is written inside the `TemporaryDirectory`, so only its position relative to the `rmtree`/`move` matters. And widening `except ValueError` to `except Exception` around the per-feature parse is untestable without contriving a parser bug; the explicit type checks now in front of it keep the blast radius small.

Also: the negative control gained a `SafeInputBaseColumn` (it claimed to cover every feature kind and covered five of six), the promotion-test helper now asserts its synthetic feature columns cannot overwrite a spine column, and the failure-modes table gets its own row rather than one cell describing two different assets' behaviour.

Full suite: 596 passed, 1 skipped. `ruff`, `ty`, `pymarkdown`, `mkdocs build --strict` all clean.

JackKelly:
🤖 Written by [Claude Code](https://claude.com/claude-code).

Merged `main` (which brought the new "Calling functions" rule — arguments are now passed by keyword at every call site this branch introduces), then ran a second independent review, on **simplicity and prose conciseness**. It was right about most things. Applied in aef47eb2.

**The guard got smaller.** The hand-rolled `isinstance` checks re-implemented a contract pydantic already owns, so `BaseForecasterConfig.model_validate` replaces them. It cannot drift from the config declaration, it chains the precise field error into the traceback, and it dissolves the bare-string-set-ified-into-characters case the previous version had to special-case by hand. Verified it rejects every shape the manual branches did: absent, `null`, bare string, non-string entries, and a non-dict `model_params`.

**The accumulator went.** Collecting every unparseable feature cost twelve lines to buy one message detail — and the remedy is the same single action ("re-train against the current vocabulary") however many names are stale. It now raises on the first, still iterating sorted so the name it reports does not vary per process.

**Two deletions.** The `meta.json`-missing branch in `fetch_model_artifacts` was unreachable and untested — `json.loads` raises the same `FileNotFoundError` naming the same path. And a parametrised test of five malformed shapes: four of them passed on `XGBoostConfig.model_validate`'s own `ValidationError`, not on this guard, so the test was pinning pydantic. The promotion-path test already covers the branch that matters, and the two promotion-refusal tests are now one parametrised test.

**Prose.** The remedy sentence had reached four copies. It stays in the error message (where the operator reads it) and in the runbook; the shell header now points at the runbook instead of repeating it. Also cut: the docstring paragraphs summarising the inherent-stability page the link already goes to, the test docstrings that restated their own names, and a comment clause claiming the promotion call site's wording is constrained by a grep it can never reach — `build_and_verify_image.sh` runs only `live_forecasts_job`, so `fetch_model_artifacts` never executes in the container.

**Mutation testing re-run after the simplification — all eight still killed:** drop the check in either call site, move it after the swap, check only the first feature, swallow the `ValidationError`, swallow the parse failure, put the tracker's name in the message, drop the source from it.

One finding declined: splitting the failure-modes table row. The reviewer offered "widen the existing row instead" as an alternative; the two outcomes genuinely differ (`live_forecasts` raising and nobody serving, versus promotion refusing and the outgoing champion continuing), so they stay as two rows — with the "Intended" cell trimmed to "Unchanged" as suggested.

Net effect on the diff: the source change is 101 lines including docstrings, down from 130. Full suite 590 passed, 1 skipped; `ruff`, `ty`, `pymarkdown`, `mkdocs build --strict` clean.

PR #550, opened 2026-08-12
Document why each defensive code path is required

JackKelly:
🤖 Written by Claude Code.

A sub-agent swept the production code for defences against states that cannot occur, and came back nearly empty: two findings, both LOW/MEDIUM confidence, both arguably legitimate. Its "considered and rejected" list was three times longer than its findings, and every entry had a real argument behind it.

The problem is that those arguments existed only in the report. So this PR writes them down at the call sites — because the next reviewer, human or agent, will otherwise reach the same "unreachable, delete it" conclusion, and will be right on the evidence the code gives them.

## What the check found

The split is sharp. The **deliberate degradation paths are documented to an unusually high standard** — `except BaseException` in `checks.py` names the `pyo3` `PanicException` reason and cites inherent-stability rule 7; the `ecmwf_ens` retry is justified across four files including the measured 3h25m republication incident; the fatal-vs-warning split in `weather_schemas.py` has a "Two things a caller must not assume" block and links to the rendered known-issues page. Nothing to do there.

**Validation that defends a package's public API, rather than a reachable state, carried no explanation at all.** That is exactly why both of the sub-agent's findings were boundary guards: with nothing to read, it reasoned from call sites, correctly found "unreachable from the one production path", and then had to talk itself back down.

## The convention

One bullet in `docs/architecture/code-style.md`, under Comments: a guard that exists for the public API rather than for a reachable state says so in a clause; a defence that only makes sense on one substrate names the substrate; repeated validation says what each call catches that the one above it did not.

## The three gaps filled

**`upsert_metadata`'s four `TimeSeriesMetadata.validate()` calls** — four calls, zero comments. `git log -S` traces the disk-read one to "Stop an unusable metadata roster wedging the hourly power ingest", then "Cut the self-healing roster rebuild" — a real incident fix whose over-built parts a previous simplicity review has already trimmed. The surviving core was load-bearing with its reasoning only in commit messages. Each call now says what it catches that the others cannot.

**`fetch_model_artifacts`' atomic directory swap** — the docstring explained the mechanism but not why atomicity applies here, when the project's stance is that torn writes are a local-root-only fault. It now says that `dest` is always local disk and is a many-file directory with no commit protocol, unlike the Delta tables.

**The two empty-input guards** (`download.py`, `geo/h3.py`) — both now marked as reusable-package input validation, naming the production path that makes them unreachable. I also replaced a "fail gracefully" comment in `download.py` with the actual reason: the Dynamical.org catalog is external and not version-pinned.

## Verification

`ruff check`, `ruff format --check`, `uv run --all-packages ty check`, `pymarkdown`, and the full `pytest` suite (613 passed, 1 skipped) are all green. Comments and one docstring only — no behaviour change.

One thing I checked rather than assumed: I had planned to justify a validate call by saying `.filter()` drops the Patito model, per the `polars-patito-gotchas` skill. It does not, for a `pt.Model.DataFrame` on this version — both `.sort()` and `.filter()` preserve the subclass. That claim is not in the diff. It may be worth checking whether the skill's wording is scoped to `pt.LazyFrame`, separately from this PR.

### DIFF
--- a/docs/architecture/code-style.md
+++ b/docs/architecture/code-style.md
@@ -139,12 +139,13 @@ Three places where a positional argument is right:
   find it.
 - **Say why a guard exists, when the reason is not "this state happens"** — validation that
   defends a reusable package's public API, rather than a state production can reach, says so in a
-  clause: `# Reusable-package input validation: unreachable from the ecmwf_ens asset, which always
-  sources h3_grid from the h3_grid_weights asset.` A defence that only makes sense on one substrate
-  names that substrate, since production data lives on S3 where a torn object write cannot happen.
-  Without the clause a reviewer traces the one production call path, finds the state impossible,
-  and proposes deleting the guard — correctly, on the evidence the code gave them. Repeated
-  validation needs the same treatment: say what each call catches that the one above it did not.
+  clause: `# Reusable-package input validation, not a reachable production state: the ecmwf_ens
+  asset always sources h3_grid from h3_grid_weights.` A defence that only makes sense on one
+  substrate names that substrate, since production data lives on S3 where a torn object write
+  cannot happen. Without the clause a reviewer traces the one production call path, finds the state
+  impossible, and proposes deleting the guard — correctly, on the evidence the code gave them.
+  Repeated validation needs the same treatment: say what each call catches that the one above it
+  did not.
 - **MkDocs-compatible constant docs** — document module-level constants with a string literal
   immediately after the assignment, not with Sphinx-style `#:` comments. This is correct:
 
--- a/docs/architecture/code-style.md
+++ b/docs/architecture/code-style.md
@@ -137,6 +137,14 @@ Three places where a positional argument is right:
   might rot is cheaper than a copy that rots invisibly. It cuts the other way too — rationale
   worth a paragraph does not belong *only* in a docstring, where no reader browsing the docs will
   find it.
+- **Say why a guard exists, when the reason is not "this state happens"** — validation that
+  defends a reusable package's public API, rather than a state production can reach, says so in a
+  clause: `# Reusable-package input validation: unreachable from the ecmwf_ens asset, which always
+  sources h3_grid from the h3_grid_weights asset.` A defence that only makes sense on one substrate
+  names that substrate, since production data lives on S3 where a torn object write cannot happen.
+  Without the clause a reviewer traces the one production call path, finds the state impossible,
+  and proposes deleting the guard — correctly, on the evidence the code gave them. Repeated
+  validation needs the same treatment: say what each call catches that the one above it did not.
 - **MkDocs-compatible constant docs** — document module-level constants with a string literal
   immediately after the assignment, not with Sphinx-style `#:` comments. This is correct:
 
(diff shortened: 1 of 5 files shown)


JackKelly:
🤖 Written by Claude Code.

Correction pushed in e8b3172a. The `compute_h3_grid_weights` comment claimed the guard protects "notebooks passing their own `h3_index`". That is wrong: no notebook calls `compute_h3_grid_weights` at all. `packages/notebooks/plot_gb_map.py` calls `compute_h3_grid_weights_for_boundary`, passing `child_h3_res=7` — the boundary wrapper, not the inner function.

Enumerating every reference, the inner function's only non-test caller is that wrapper, one function above it in the same module, which already raises on an empty cell list. The comment now says that, and points at `test_h3.py` where the empty-input guard is exercised directly.

I re-checked the sibling claim in `download.py` at the same time and it holds: `open_ecmwf_ens_run`'s only non-test caller is `assets.py:288`, and the `ecmwf_ens` asset reads `h3_grid` from `settings.h3_grid_weights_path`, which only the `h3_grid_weights` asset writes.

JackKelly:
🤖 Written by Claude Code.

Adversarial review (Sonnet 5) run over the diff; fixes applied in 64557590.

**One factual claim was wrong.** The comment on the third `TimeSeriesMetadata.validate` call said the diagonal concat "catches a null landing in a field the snapshot itself carried non-null". It cannot: diagonal concat null-fills a column only for rows from the frame that *lacks* it, never overwriting a value already present. Reproduced directly to confirm. Digging further, all four `allow_missing` fields on `TimeSeriesMetadata` are nullable (`str | None` / `float | None`), so a null arriving that way is legal and that call cannot fail the way any version of the comment implied.

The comment now says what the call actually does — a shape check on the first frame carrying the union of both inputs' columns — and marks it the weakest of the four, and the first to reconsider if these get trimmed. I also dropped the "none of the four is redundant" claim from the first comment, which I could not defend.

**Follow-up worth its own look, out of scope here:** given that #1 and #2 already validate both inputs, and that the concat cannot introduce an illegal value, validation #3 may be genuinely redundant. #4 is defensible as the last gate before overwrite, but on value grounds it is also weak. This is a behaviour change so it is not in this comments-only PR, but it is a real simplicity finding and the code now points a reader at it.

**Also fixed:** `dest` is local disk *by convention* — `production_model_path` is an unvalidated `str` field with no scheme restriction (unlike `nged_s3_bucket_url`, which has a `field_validator`), so "always" was overstated. The containment note ("the asset records `metadata_upsert_failed`") was attached to the second guard as though it distinguished it, but `assets.py` wraps the whole `upsert_metadata` call, so it is true of all four; rescoped. Plus two wording nits and a re-wrap to stay under 100 characters.

**Verified correct and unchanged:** the `geo/h3.py` caller enumeration, the `download.py` → `h3_grid_weights` chain, the Dynamical.org not-version-pinned claim, and the first and fourth `storage.py` comments.

Green: `ruff check`, `ruff format --check`, `ty`, `pymarkdown`, full `pytest` (613 passed, 1 skipped).

PR #551, opened 2026-08-12
Correct the pt.LazyFrame.filter() gotcha: a typing gap, not a runtime loss

JackKelly:
🤖 Written by Claude Code.

The `polars-patito-gotchas` skill claimed `.filter()` "drops the Patito subclass". It doesn't, and the claim was unscoped enough to read as applying to `pt.DataFrame` as well as `pt.LazyFrame`.

## What's actually true

**At runtime the model survives.** On patito 0.8.6 / polars 1.43.2, `.filter()`, `.sort()`, `.select()`, `.with_columns()`, `.head()` and `.unique()` on a `pt.LazyFrame` *and* on a `pt.DataFrame` all return the model-bearing subclass with `.model` still set. Nothing is dropped and nothing needs restoring.

**The trap is a type-annotation gap.** `ty` types `scan.filter(...)` on a `scan: pt.LazyFrame[Schema]` as `polars.lazyframe.frame.LazyFrame`, so `scan = scan.filter(...)` fails:

```text
error[invalid-assignment]: Object of type `polars.lazyframe.frame.LazyFrame` is not assignable to `patito.polars.LazyFrame[PowerForecast]`
```

**`pt.DataFrame` is unaffected, in both senses.** `patito/polars.py` gives `patito.polars.DataFrame` a `# --- Type annotation overrides ---` block re-declaring `filter`/`select`/`with_columns` as `(self: DF) -> DF`; `patito.polars.LazyFrame` has no such block and inherits Polars' own annotations. So `df.filter(...)` on a `pt.DataFrame[S]` stays `pt.DataFrame[S]` to `ty` and needs no workaround at all.

## What changed

- The skill section, rewritten with the real mechanism, the actual `ty` error text, and an explicit "`pt.DataFrame` is unaffected" note. The workaround code is unchanged — it was always correct, it just wasn't doing what the prose said.
- The skill's front-matter description and intro paragraph. The intro said every trap "produces no error at the point of the mistake", which sat awkwardly above a section whose whole content is the error you get; it now says none of them *points at itself*, which covers both kinds.
- `docs/architecture/code-style.md`, where the Patito friction budget summarises the five gotchas by name.
- Three comments in `cv_assets.py` that repeated the wrong wording verbatim.

No behaviour change; comments, docs and one skill only.

## How this was found

While documenting boundary validation in #550 I nearly repeated the `.filter()` claim in a code comment, tested it first, and found it false for `pt.DataFrame`. This PR is the follow-up: I checked runtime behaviour across six methods on both frame types, then checked what `ty` infers for each, then read `patito/polars.py` to find why the two disagree.

## Verification

`ruff check`, `ruff format --check`, `ty check` (all packages), `pymarkdown` and `pytest` (614 passed, 1 skipped) all green.

### DIFF
--- a/.claude/skills/polars-patito-gotchas/SKILL.md
+++ b/.claude/skills/polars-patito-gotchas/SKILL.md
@@ -3,18 +3,19 @@ name: polars-patito-gotchas
 description: >-
   Five Patito/Polars/Delta traps that fail silently rather than raising: cross-model LazyFrame
   joins, a `{column: dtype}` cast swallowed on a model-bearing frame, `ge`/`le` ignored on a
-  datetime field, `.filter()` dropping the Patito subclass, and dictionary-encoded columns
-  blocking Delta predicate pushdown. Load before writing Polars/Patito code that joins, casts,
-  filters a `pt.LazyFrame`, declares a Patito field, or reads/writes Delta — or when a
-  `validate()` dtype error, a `.join()` TypeError, or an over-reading Delta scan looks
-  inexplicable.
+  datetime field, `pt.LazyFrame` methods typed as plain `pl.LazyFrame`, and dictionary-encoded
+  columns blocking Delta predicate pushdown. Load before writing Polars/Patito code that joins,
+  casts, filters a `pt.LazyFrame`, declares a Patito field, or reads/writes Delta — or when a
+  `validate()` dtype error, a `.join()` TypeError, an `invalid-assignment` on a filtered scan, or
+  an over-reading Delta scan looks inexplicable.
 ---
 
 # Patito + Polars + Delta gotchas
 
-Every trap below produces **no error at the point of the mistake**. They surface later as a
-confusing `validate()` failure, a `ty` complaint several lines away, or a query that quietly
-reads the whole table. That is why they are written down.
+**None of the traps below points at itself.** Most produce no error where the mistake is, surfacing
+later as a confusing `validate()` failure or a query that quietly reads the whole table; the rest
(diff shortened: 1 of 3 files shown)


JackKelly:
🤖 Written by Claude Code.

**Checked whether a newer `ty` fixes this. It doesn't, and it can't — `ty` is not wrong here.**

`polars/lazyframe/frame.py` annotates `filter`, `sort`, `select`, `with_columns`, `head`, `unique`, `drop` and `rename` as returning `LazyFrame`, not `Self`. Given that annotation, inferring the base class is the *correct* result for any conforming checker, and there is no `ty` release that changes it.

Two pieces of evidence:

- Reproduced on `ty` 0.0.70 (latest on PyPI) as well as the 0.0.69 currently installed — byte-identical `invalid-assignment` diagnostic.
- `pyright`, a completely independent implementation, reports the same thing: `error: Type "LazyFrame" is not assignable to declared type "LazyFrame[PowerForecast]" (reportAssignmentType)`.

The fix has to come from upstream: either Polars switching those return annotations to `Self`, or Patito giving its `LazyFrame` the `# --- Type annotation overrides ---` block its `DataFrame` already has. That second one is the smaller change and it is exactly why `pt.DataFrame` needs no workaround today.

Pushed as bd34cec5, which adds that to the skill section so the next person doesn't re-run this experiment. No version numbers went into the skill text — the claim holds for any checker version, and a pinned number would just go stale.

This PR touches no `ty` pin, so it will not conflict with the `ty` upgrade branch.

PR #592, opened 2026-08-13
Document how thin a partially-covered H3 cell can get

### LINKED ISSUE #506: Report the contributing-weight fraction from the H3 NWP aggregation
*Filed by Claude Code, at Jack's direction, from the planning of #496 + #477.*

Once the H3 aggregation renormalises over the non-null contributing grid points (#496), each
`(h3_index, variable)` has a **contributing weight**: the sum of the `proportion`s of the grid
points that actually supplied a value. It is 1.0 for a fully-covered cell and less than 1.0 when
some of the cell's grid points were null or absent. #496 raises the question of whether to surface
that number and deliberately leaves it open; this issue is that question.

## Why it might be worth having

Renormalisation makes a partially-covered cell yield a value rather than a null. That is the point
of #496 — but it also means a cell backed by one surviving grid point out of five is now
indistinguishable, in the data, from one backed by all five. The contributing weight is the only
thing that can tell them apart.

Emitting it is also the alternative #496 prefers to a minimum-coverage *threshold*: "better, and
more in keeping with the repo's dislike of magic thresholds — emitting the contributing-weight
fraction so the quality check can report it". No floor should be introduced in either case.

## What the shape of the answer is not

**Not per-row columns in the `Nwp` contract.** That is 12 extra `Float32` columns on a table that
is already ~5.9 billion rows, carrying 1.0 in the overwhelming majority of them. Rejected on cost
before the design starts.

**Not a per-cell number, either — the weight is per `(cell, variable)`.** At lead-0 all three
de-accumulated variables are null at every grid point, so their contributing weight is 0 in every
cell of every run. That is correct and expected, but it means anyone reading a single number as
"how much of this cell arrived" would be badly misled.

## What to decide

- Where it lives: a run-level summary on `NwpQualityReport`, materialisation metadata on the
  `ecmwf_ens` asset, or something else. A run-level summary changes
  `convert_nwp_xarray_dataset_to_polars_dataframe`'s return type and threads through the asset, so
  it is a real design change rather than a line in a diff.
- What is summarised: a count of partially-covered `(cell, variable)` pairs, a distribution, or a
  worst case.
- Whether it is worth having *at all* given the sibling issue below, which may be the better answer
  to the underlying question.

## Relationship to the sibling issue

This and "move the upstream ECMWF-corruption signal to grid-point level" are two answers to the
same underlying question — how do we still see upstream corruption once renormalisation absorbs
it — approached from opposite ends. The grid-point-level signal measures the corruption where it
actually lives, before our aggregation touches it; this issue measures what our aggregation did
about it. They are complementary rather than alternatives, but the grid-point one is the more
directly useful of the two and should be settled first.

Depends on #496 landing.

JackKelly:
*Written by Claude Code, from the investigation recorded on #506.*

Docs only — two paragraphs added to [Known ECMWF ENS data-quality issues](https://openclimatefix.github.io/nged-substation-forecast/architecture/ecmwf-ens-known-issues/#spatial-aggregation-is-where-a-grid-points-null-is-resolved). No code changes.

## The gap

That page explains that the H3 aggregation renormalises each cell over the grid points that supplied a value, so a corrupt grid point costs only its own share of its cell. What it never says is how far that can go. A cell reduced to a single surviving point still yields a value, and it is stored exactly like a fully-covered one: the `Nwp` contract has no column for the contributing weight, and `_aggregate_grid_points_to_h3_cells` drops it once the division is done. A reader finishes the section knowing a null cell means "no point contributed" without knowing that a *non-null* cell might rest on almost nothing.

## What was added

**The bound**, which is a fixed property of the geometry rather than a per-run number. `compute_h3_grid_weights_for_boundary` defaults `child_h3_res` to `h3_res + 2`, so each cell is sampled with 49 H3 children and every `proportion` is a multiple of 1/49 — verified against `data/h3_grid_weights.parquet`, where the maximum deviation from an integer after multiplying by 49 is 1.9e-06. The least a single grid point can carry is therefore 2.0% of its cell. 572 of the V1 grid's 1671 cells contain a point that thin, and 859 contain one under 10%.

**Why no per-run metric reports the weight**, so the question does not get reopened by the next person who notices the gap. Measured on the real V1 geometry by nulling *k* grid points and computing each candidate summary against `n_null_nwp_grid_points`, 20 trials per point:

| | scattered | clustered |
|---|---|---|
| partially-covered cells | 4.7 × `n_null_nwp_grid_points` | saturates at ~55 |
| total weight deficit | 1.68 × `n_null_nwp_grid_points` | saturates at ~27.7 |
| thinnest surviving weight | pins to the 1/49 floor | pins to the 1/49 floor |

The scattered multipliers are arithmetic rather than coincidence: 1.68 is the mean total `proportion` a grid point carries across the 4.92 cells it feeds. Under clustered corruption — the shape [the page already says](https://openclimatefix.github.io/nged-substation-forecast/architecture/ecmwf-ens-known-issues/#nulls-in-the-de-accumulated-variables-tolerated) it takes in practice — the interior of a corrupt patch goes null and is counted by `n_null_h3_cells`, leaving only the rim partially covered, so both counts stop moving while the damage keeps growing. For scale, the worst archived run (2025-06-04 00Z, 339 null cells) sits inside that saturated regime.

So every available summary either duplicates a number `nwp_has_no_unexpected_nulls` already publishes or goes blind exactly when it would matter. That is the finding that closed #506 as not planned, and this PR is the part of it worth keeping.

## Scope

Deliberately no code. The measurements above were made in a throwaway script against `data/h3_grid_weights.parquet`; nothing from it is committed, because the numbers are properties of a stored artefact that anyone can recompute, and a test asserting them would pin the docs to one particular H3 grid.

The `nwp_grid_point` / `h3_cell` metadata keys named in the new prose are the ones #573 introduced, so this sits on top of that work rather than changing it.

## Review

Sized as simple: prose only, one file, no contract, no asset, no production path. A single adversarial sub-agent reviewed the new text for factual accuracy against the code and data.

## Verification

`pymarkdown scan -r docs README.md CLAUDE.md packages/*/README.md` and `mkdocs build --strict` both clean, and the rendered HTML under `site/` was read to confirm the two paragraphs render as paragraphs — no stray headings, code spans intact. `ruff`, `ty` and `pytest` were not run: no Python file is touched, and no test reads `docs/`.

### DIFF
--- a/docs/architecture/ecmwf-ens-known-issues.md
+++ b/docs/architecture/ecmwf-ens-known-issues.md
@@ -116,6 +116,23 @@ the upstream scatter documented above lands outside the small GB box we download
 therefore a correctness fix — the estimator was wrong, and a wholly-uncovered cell was a false
 zero — rather than a change that recovers much data.
 
+A surviving cell can rest on very little of its own area. `compute_h3_grid_weights_for_boundary`
+samples each cell with H3 children two resolutions finer — 49 of them — so every `proportion` is a
+multiple of 1/49, and the least a single grid point can carry is 2.0% of its cell. 572 of the V1
+grid's 1671 cells contain a point that thin, and 859 contain one under 10%. A cell reduced to one of
+those points is stored exactly like a fully-covered one: the `Nwp` contract has no column for the
+contributing weight, and `_aggregate_grid_points_to_h3_cells` drops it once the division is done.
+
+No per-run metric reports that weight either, because every available summary of it either
+duplicates a number `nwp_has_no_unexpected_nulls` already publishes or goes blind exactly when it
+would matter. Where the corruption is scattered, the count of partially-covered cells runs at 4.7
+times `n_null_nwp_grid_points` and the total weight deficit at 1.68 times it — the deficit being the
+mean total `proportion` a grid point carries across the 4.92 cells it feeds. Where it is clustered,
+which is the shape it takes in practice, both saturate: the interior of a corrupt patch goes null
+and is counted by `n_null_h3_cells`, so only the rim stays partially covered and the two numbers
+stop moving while the damage grows. The thinnest surviving weight is no better a summary, since it
+pins to the 1/49 floor on any run corrupt enough to ask about.
+
 A cell where *no* point contributed is a different case, and it yields **null**, never `0.0`. That
 distinction is the whole reason the contributing weight is computed rather than assumed: Polars
 sums an all-null group to `0.0`, which for weather is a physically plausible, in-bounds lie


JackKelly:
Superseded: the body registered an unintended closing link to #506, which cannot be edited out. Reopening from the same branch with corrected wording.

PR #595, opened 2026-08-13
Runbook: vacuum cannot reclaim the superseded NWP parquet files

### LINKED ISSUE #593: DeltaTable.vacuum() silently deletes nothing on the nwp table: vacuum re-encodes the already-encoded init_time partition path
🤖 Written by Claude Code.

`DeltaTable.vacuum()` — that is, lite mode, the default — cannot delete a single file from the `nwp` Delta table. It reports success, returns the paths it claims to have deleted, and writes `VACUUM START` / `VACUUM END` commits to the transaction log, while every file stays on disk. `vacuum(full=True)` does work, by a different code path; see "The workaround" below.

## What happened

After #589 made `write_nwp` replace its partition, re-materialising the `2026-08-13` partition left the superseded pair of parquet files on disk, as expected — nothing runs `vacuum`. Running it by hand to reclaim them:

```python
dt.vacuum(retention_hours=0, dry_run=False, enforce_retention_duration=False)
```

returned both files and moved the table from version 865 to 867. The table grew by 584 bytes (the two vacuum commits) and both files were still there. They had to be removed with `rm`, after cross-checking them against the current version's add actions.

The paths vacuum reported are **double-encoded**:

```text
init_time=2026-08-13%252000%253A00%253A00.000000/part-00000-ee13a2b2-….parquet
```

The directory on disk is `init_time=2026-08-13%2000%3A00%3A00.000000`.

## Why

Two faults in delta-rs 1.6.2 combine. Either alone would be survivable.

**1. Vacuum re-encodes a path that is already encoded.** Because `init_time` is a `Timestamp`, its Delta partition value is `2026-08-13 00:00:00.000000`, whose space and colons are percent-encoded into the Hive directory name — so the directory name itself contains literal `%` characters, and the transaction log stores that name URL-encoded, turning each `%` into `%25`. **The `%2520` in our log is correct**, and decoding it once gives the real directory name, which is why every read works and why this was invisible until now. Active files are turned into object-store paths by `LogicalFileView::object_store_path` ([`iterators.rs:140`](https://github.com/delta-io/delta-rs/blob/main/crates/core/src/kernel/snapshot/iterators.rs#L140)), which tries `Path::parse` first and so preserves that encoding. Vacuum builds tombstone paths with bare `Path::from` ([`vacuum.rs:329`](https://github.com/delta-io/delta-rs/blob/main/crates/core/src/operations/vacuum.rs#L329)), which percent-encodes, and object_store's invalid-character set includes `%`. Every `%` gains another `%25`, and the delete targets a path that does not exist.

**2. A `NotFound` while deleting is reported as a deletion.** [`vacuum.rs:524-530`](https://github.com/delta-io/delta-rs/blob/main/crates/core/src/operations/vacuum.rs#L524) maps `Err(NotFound)` to `Ok(path)`, so the missed delete lands in the returned list and in `numDeletedFiles` on the `VACUUM END` commit. This is what turns fault 1 from a no-op into a silent one.

The tell is in the return value: a real deletion returns a relative object-store path, whereas the `NotFound` arm returns the error's path, which `LocalFileSystem` renders absolute.

## Minimal repro

Four tables, from two independent variables: the partition column's type (`Timestamp`, whose directory name needs escaping, versus a `String` control) and the vacuum mode. Each is written twice to the same partition (the second write an `overwrite` with a predicate), then vacuumed at zero retention. On `deltalake` 1.6.2 and `polars` 1.43.2:

```text
=== Timestamp partition column, vacuum(full=False) ===
  on-disk directory name:  part=2026-08-13%2000%3A00%3A00.000000
  log add: part=2026-08-13%252000%253A00%253A00.000000/…
  parquet files before vacuum: 2
  vacuum reported deleting:    1
    reported: /tmp/…/ts_lite/part=2026-08-13%252000%253A00%253A00.000000/part-00000-….parquet
    ^ absolute (i.e. a NotFound error path): True
  >>> files actually removed:  0

=== Timestamp partition column, vacuum(full=True) ===
  parquet files before vacuum: 2
  vacuum reported deleting:    2
    reported: /tmp/…/ts_full/part=2026-08-13%252000%253A00%253A00.000000/part-00000-b3be7dc3-….parquet
    ^ absolute (i.e. a NotFound error path): True
    reported: part=2026-08-13%2000%3A00%3A00.000000/part-00000-b3be7dc3-….parquet
    ^ absolute (i.e. a NotFound error path): False
  >>> files actually removed:  1

=== String partition column, vacuum(full=False) ===
  on-disk directory name:  part=2026-08-13T00Z
  parquet files before vacuum: 2
  vacuum reported deleting:    1
    reported: part=2026-08-13T00Z/part-00000-….parquet
    ^ absolute (i.e. a NotFound error path): False
  >>> files actually removed:  1
```

The `String` control passes, so this is not vacuum being broken in general — it is specific to a partition value whose directory name needs percent-encoding. The `full=True` case is the whole bug in miniature: the same file is reported twice, once as the over-encoded absolute `NotFound` path and once as the real relative path, so `numDeletedFiles` says 2 for one file.

## The workaround

`vacuum(full=True)` reclaims the space, with no change to how we name partition directories. Full mode lists actual storage and deletes at `obj_meta.location`, which is correctly encoded.

It works by a different route than intended, so it is worth knowing what it does and does not protect. Delta-rs never recognises our superseded file as a tombstone at all — the tombstone path set is built with the same broken `Path::from` ([`vacuum.rs:620`](https://github.com/delta-io/delta-rs/blob/main/crates/core/src/operations/vacuum.rs#L620)) — so it classifies the file as an untracked stale orphan and deletes it on that basis. Two protections still hold, both because they use `object_store_path()`: files referenced by the current version (`valid_files`, `vacuum.rs:513`) and files pinned by `keep_versions` (`vacuum.rs:274`).

The protection that does **not** hold is "keep a recently-tombstoned file for the retention window". It is replaced by the file's physical age (`vacuum.rs:376`), and our superseded file was written when the run first landed — possibly weeks before it was tombstoned. So `full=True` can sever time travel to the pre-replacement version immediately, rather than after `deletedFileRetentionDuration`. The `nwp` table sets no `configuration`, so that default is 7 days.

The practical consequence: do not run it while an in-place rewrite is reading from a pinned pre-rewrite version, because the usual grace period that would forgive the mistiming is not there.

## Why it matters now, and did not before

`write_nwp` used to be append-only, so nothing ever superseded a file and there was nothing to reclaim. Since #589 a partition replace leaves a full dead copy of the run behind — 137 MiB per re-materialisation of a V1 partition. Correctness is unaffected: every reader goes through the transaction log, and the live version's files are never the ones at risk.

The silence is the worse half. Anyone reading lite vacuum's return value would believe space had been reclaimed and time travel severed, when neither happened.

## What to do

1. ~~**Report it upstream to delta-rs.**~~ **Done** — filed as [delta-io/delta-rs#4646](https://github.com/delta-io/delta-rs/issues/4646) (the vacuum bug) and [delta-io/delta-rs#4647](https://github.com/delta-io/delta-rs/issues/4647) (an opt-in choice of partition directory format, which would remove the precondition). The upstream report establishes two things this issue did not: the bug is a regression introduced in `deltalake` 1.5.0 by delta-io/delta-rs#4227, and it is not specific to `Timestamp` partitions — any partition directory name containing a character in object_store's `INVALID` set hits it, including values with no `%` in the name at all.
2. **Decide whether to reopen #294 and partition on a column whose directory name needs no escaping.** #294 worked out that the serialization format is fixed by the Delta protocol and that the only lever is to stop partitioning on the raw `Timestamp`; it was closed as not-planned because it was judged purely cosmetic, and this issue is the evidence that it is not.

    **The partition key has to keep the run hour**, so that we retain the option of ingesting an NWP that updates more than once a day. That rules out a plain `Date`, which is otherwise the tidiest answer. Measured on `deltalake` 1.6.2 for an 06Z run, these all produce a directory name needing no escaping:

    | Scheme | Directory |
    | --- | --- |
    | `String` | `init_time_utc=2026-08-13T06Z` |
    | `Date` + `Int8` hour | `init_date=2026-08-13/init_hour=6` |
    | `Int64` packed | `init_yyyymmddhh=2026081306` |
    | `Int64` epoch | `init_epoch=1786600800` |

    A derived column is unavoidable: every `Datetime` writes `%2000%3A00%3A00.000000` whatever its time unit or time zone, no Delta type serialises an hour without a colon, and no setting changes it — the partition value's string form is fixed by the protocol, and the escaping into a directory name is delta-rs' own, added deliberately so that colons do not break Windows (delta-io/delta-rs#2382).

    `Date` + `Int8` hour is the one that needs no string at all: two honestly-typed columns, both readable, hierarchical pruning, and any sub-daily cadence works. The packed `Int64` keeps it to a single column but is an integer pretending to be a timestamp, which trades the same honesty a string does while being harder to read in a filter.

    `init_time` stays a real `Timestamp` column for the primary key under any of these, and the partition columns cost nothing on disk — verified by reading the parquet schema of a partitioned table, which holds only the non-partition columns. The real cost is that a reader filtering on `init_time` alone stops pruning. NWP reads already funnel through `Nwp.scan_delta` and `weather_utils`, so that translation could live in one place rather than in every caller.

    This is the *thorough* fix rather than the *necessary* one, since `full=True` reclaims the space today, and the migration cost is real: a full rewrite of the 93 GB local table and, eventually, the S3 one.
3. **Feed the `full=True` caveats into #357**, which is where a scheduled vacuum job would be designed. Its "pick a retention window deliberately" step has to account for the retention window not being honoured for these files.

## Relationship to #476 / #580 / #589

None of them caused this; #589 only made it reachable, by creating superseded files for the first time.

JackKelly:
🤖 Written by Claude Code.

One sentence in the NWP runbook, corrected. It does not fix #593 and deliberately carries no closing keyword — it only stops the runbook implying a command works when it does not.

#589 added a bullet saying the superseded parquet files stay on disk "and nothing here runs `vacuum`". True, but it reads as though running `vacuum` would clear them. On this table it does not: `DeltaTable.vacuum` reports deleting the files, writes `VACUUM START` / `VACUUM END` commits, and removes nothing, because `init_time`'s Hive directory name is percent-encoded and vacuum deletes at a double-encoded path that does not exist. Found by running it against the real table after re-materialising the `2026-08-13` partition; root cause and a minimal repro with a passing `String`-partitioned control are in #593.

The bullet now says vacuum will not clear them, why, and what reclaiming the space actually takes until #593 is resolved. The dead-row figure gains its size in MiB (~137 MiB per re-materialised V1 partition), since that is the number an operator watching disk usage wants.

Docs-only. `pymarkdown scan` and `mkdocs build --strict` are clean, and the rendered HTML was read to confirm the wrapped link did not turn into a heading and the list item survived intact.

### DIFF
--- a/docs/live_service/operations.md
+++ b/docs/live_service/operations.md
@@ -320,9 +320,16 @@ Two things a re-run does cost:
   running.** The two writes contend and the loser fails with delta-rs' `CommitFailedError`. That
   costs a run, not the table, and re-running afterwards is safe. Partitions for *different* dates
   do not contend at all, so a backfill alongside the daily schedule is fine.
-- **The superseded rows stay on disk.** Delta marks the old parquet files as removed rather than
-  deleting them, and nothing here runs `vacuum`, so replacing a V1 partition leaves ~7.24M dead
-  rows behind. Reads are unaffected — every reader goes through the transaction log.
+- **The superseded rows stay on disk, and `vacuum` will not clear them.** Delta marks the old
+  parquet files as removed rather than deleting them, so replacing a V1 partition leaves ~7.24M
+  dead rows — about 137 MiB — behind. `DeltaTable.vacuum` reports deleting those files and does
+  not: the `init_time` partition directory's name is percent-encoded, and vacuum deletes at a
+  double-encoded path that does not exist, so it removes nothing while writing vacuum commits
+  saying it did
+  ([issue #593](https://github.com/openclimatefix/nged-substation-forecast/issues/593)). Reads are
+  unaffected — every reader goes through the transaction log. Until that is fixed, reclaiming the
+  space means listing the current version's add actions, diffing them against what is on disk, and
+  deleting the orphans by hand.
 
 Every materialisation whose completeness assessment succeeded also publishes `n_ensemble_members`,
 `n_valid_times`, `n_h3_cells` and the `valid_time` range as metadata, so the Dagster UI timeline


JackKelly:
Closing and reopening from the same branch: the original body said "does not fix #593", and GitHub's link parser ignored the negation and registered a closing link to that issue. Closing links are sticky and cannot be edited out, so this PR would have closed #593 on merge. Replaced by a PR with a clean body; no code or docs change between them.

PR #597, opened 2026-08-13
Categorise Sentry events, and stop a transient S3 blip becoming one

### LINKED ISSUE #488: Categorise Sentry events
Eg to distinguish between "upsteam data issue" versus "bug in our code". See this discussion: https://claude.ai/share/c5c567aa-180f-4728-a99d-9356e8691f79

JackKelly:
*Written by Claude Code, at Jack's direction.*

Closes #488.

**Sizing and review.** The issue was sized **complex** — it changes the production failure behaviour of the hourly ingest, which is a degradation rule, and more than one design defensibly satisfies it. It therefore got a written plan, both adversarial plan reviews, and it is getting both adversarial diff reviews. The plan is in `plans/claude-categorise-sentry-events.md` in this diff, including what each review changed and what was rejected; it will be deleted into this body at ship time.

## What was wrong

The issue's comment names three classes of Sentry event. Only one of them needed code:

1. **A transient object-store blip in `power_time_series_and_metadata`** failed the hourly run and reported exactly like a genuine bug, even though no data was lost and the next hourly run back-fills.
2. **A genuine bug in our code** — what the failure hook is for, and already correct.
3. **"A warning path could not evaluate its inputs"** — already tagged `asset_check`. As the comment suspected, this class was free; it gets a docs line and no code.

A fourth problem surfaced while verifying the first: an exhausted in-band retry reaches Sentry titled and grouped as `RetryRequested` rather than as the fault that caused it. That is live today for `ecmwf_ens` — a run that never publishes groups as `RetryRequested`, hiding `NwpRunNotYetAvailable`.

## What changed

- **`power_time_series_and_metadata` retries the NGED S3 read in-band**, twice, seconds apart, so a blip never becomes an event. The budget is deliberately tiny: nothing is at risk while we wait, because the asset re-lists the bucket from scratch on each attempt and the next hourly run back-fills anyway, so the retry only has to cover an instantaneous glitch. A persistent outage still exhausts the budget, fails and reports.
- **`sentry_capture_failure` unwraps a plain `RetryRequested` to its cause**, so an exhausted retry is titled and grouped by the real fault. Dagster's `HookContext.op_exception` already unwraps its own `RetryRequestedFromPolicy`; this extends that one class up the hierarchy.
- **`sentry_capture_failure` drops deliberate exits** — `KeyboardInterrupt`, `SystemExit`, `DagsterExecutionInterruptedError`. Cancelling a run is an operator's decision, not a fault. `SystemExit` reaches the hook today; the other two are defence in depth.
- **The failure hook tags `fault_category:run_failed`.** Only this sender gets a tag, because the other three already carry a mark: `degraded_asset`, `asset_check`, and `level="warning"` plus a stable fingerprint. What the failure hook lacked was any positive marker at all, and it is the one class worth alerting a human about — so it is the one that must not be identified by elimination. A rule phrased "error level, and neither degradation tag is set" is correct today and misclassifies silently the day a fifth sender is added.

## Why not a Dagster `RetryPolicy`, which the issue suggested

Because it makes a cancelled run finish **green**, having written its data. Measured against Dagster 1.13.17:

```
ONE-SHOT INTERRUPT + POLICY: run_success=True attempts=2 side_effects=['WROTE DATA']
```

`dagster._core.execution.plan.utils` converts an interrupt into a retry request whenever the op carries a policy, and Dagster delivers a termination as exactly one interrupt. That interrupt is consumed by the conversion, the step restarts, no second signal arrives, and the run completes. It directly defeats the `raise  # A cancelled run must cancel.` line this repo maintains deliberately in that very asset.

A `RetryPolicy` also retries the *whole* asset body, so a deterministic bug occurring after the Delta append would be retried, dedupe to a no-op on the second attempt, and turn a real failure into a green run — every hour, forever, with nothing in Sentry. Raising `RetryRequested` from inside a guard that re-raises the three interrupt types avoids both, keeps cancellation immediate, aims the retry at the upstream read alone, and matches the idiom `ecmwf_ens` already uses.

This finding is recorded in `docs/architecture/production-deployment.md`, so the next person reaching for a `RetryPolicy` finds the evidence rather than the trap.

## Why no "upstream exception types" list

The issue's comment predicted such a list would drift, and it is right, so there is none. Each compiled extension the ingest path reads through — obstore, delta-rs, polars — defines its own exception classes, and a pyo3 panic is not an `Exception` at all. The guard therefore catches `BaseException` and names what must *propagate*, which is the same reasoning `checks.py::power_data_is_fresh` already carries in a comment. Sentry also already groups by exception type and puts it in the issue title, so a tag restating it would add nothing.

## Tests

Ten new tests. Eight fail on `main`; the three that pass there are guards against a wrong implementation rather than tests of a behaviour change, and are labelled as such:

| Test | Fails on `main`? |
|---|---|
| `test_failure_hook_reports_the_cause_of_a_retry_requested` | yes |
| `test_failure_hook_ignores_a_deliberate_exit` (4 cases) | yes |
| `test_failure_hook_tags_the_fault_category` | yes |
| `test_power_time_series_and_metadata_retries_a_transient_upstream_failure` | yes |
| `test_power_time_series_and_metadata_gives_up_after_its_retry_budget` | yes |
| `test_failure_hook_captures_a_retry_requested_with_no_cause` | no — guards a `None` dereference |
| `test_power_time_series_and_metadata_does_not_retry_a_cancelled_run` | no — regression guard for the defect above |
| `test_power_time_series_and_metadata_handles_no_new_data` (extended) | no — pins `NoNewData` ahead of the retry guard |

The retry test asserts the written data, not just the call count: a second attempt re-lists and re-downloads everything, so it also pins down that re-running the ingest cannot duplicate rows.

## Verification

`ruff check`, `ruff format --check`, `ty check`, `pytest` (687 passed, 1 skipped), `pymarkdown scan` and `mkdocs build --strict` are all green. The rendered HTML under `site/` was read for the three docs pages that changed, per the `mkdocs-authoring` rule. Suite runtime went from ~40 s to ~47 s: the two retry tests sleep through their budgets by design.


JackKelly:
*Written by Claude Code.* Both adversarial diff reviews have now run, each by a fresh sub-agent with no access to the implementer's reasoning. Summary of what they changed, and what I rejected.

## Review 1 — correctness, and cutting the change down

It confirmed every load-bearing claim (the retry semantics, the `NoNewData` ordering, the no-op-without-DSN behaviour, the "8 of 10 tests fail on main" table, and that the restructure hides no behaviour change), and found the defects in the prose and tests rather than the mechanism.

**Accepted:**

- **`test_failure_hook_tags_the_fault_category` could not fail.** It asserted `{FAULT_CATEGORY_TAG: RUN_FAILED_FAULT_CATEGORY}` against the module's own constants — true for any pair of values, while the production alert rule depends on the exact strings. It now asserts the literals. Verified by mutation: renaming the constant now turns it red.
- **6 s of suite wall clock bought nothing.** Neither retry test asserts on timing, so they now monkeypatch the delay to 0. Those two files went from 10.4 s to 4.7 s.
- **A comment described a mechanism that cannot occur** — it said our own guards produce a wrapped cancellation. None of them do; they all re-raise. The only producer is Dagster itself, for an op carrying a `RetryPolicy`. Rewritten to say so.
- **Which interrupt shape actually reaches the hook.** Dagster re-raises `KeyboardInterrupt` and `DagsterExecutionInterruptedError` ahead of it; `SystemExit` is the one that arrives. The docstring and docs now say that rather than implying all three were previously reported.
- Helper consistency, the duplicated tag vocabulary between `operations.md` and `sentry.md`, an over-long `RetryPolicy` rationale, and a metaphor in `aws.md` that the repo's prose rule forbids.

**Rejected:**

- **"Inline `FAULT_CATEGORY_TAG` / `RUN_FAILED_FAULT_CATEGORY`, they have one use each."** `POWER_DATA_STALE_FINGERPRINT` in the same module is also a single-use documented `Final`, so this is house style, and these two strings are a contract with a Sentry console rule. Their docstrings were cut back instead.
- **"Cut the two unreachable cases from the interrupt parametrize."** They cost one line each and pin the guard's stated contract against Dagster changing its routing. The docstring now says which case is reachable.

## Review 2 — mutation testing

30 mutations, 22 killed, 8 survived. Three were real gaps, all now closed and each confirmed to turn the suite red:

1. **Dropping `from exc` on the guard's raise.** The highest-value survivor: it silently disables the feature this PR exists to deliver, because the hook unwraps `__cause__` and without `from exc` Python sets only `__context__`. The unwrap was tested against a hand-built fixture, so nothing connected the raising side to the unwrapping side. The budget test now asserts on the `__cause__` of what the asset actually raised.
2. **Extending the retry guard over the writes.** This is exactly the failure mode the comment there warns about — a bug after the Delta append is retried, the second attempt's `select_new_rows` dedupes it to a no-op, and a real failure lands as a green run. The boundary was entirely unpinned; there is now a test that a post-write failure runs once and fails.
3. **`_POWER_INGEST_MAX_RETRIES = 99`.** The budget test computed its expectation from the constant, so it was self-referential. It asserts the literal now — a mutant that re-listed and re-downloaded NGED's bucket 100 times per failing hour was reporting green.

The surviving mutations I did **not** close, with reasons:

- **Dropping `KeyboardInterrupt` / `SystemExit` from the *asset's* re-raise union** — closed anyway, by parametrising the cancellation test over all three types. Cheap and symmetric with the hook's test.
- **Anything about the retry *delay*** (dropping `seconds_to_wait`, changing its value). Asserting on a real sleep buys a slow test for a parameter with no correctness content — the only wrong values are ones so long the hourly schedule laps them, which no unit test can police.
- **Deleting the `context.log.warning`.** UI-only diagnostic text; pinning log strings is brittle.

## State

`ruff`, `ruff format --check`, `ty check`, `pymarkdown` and `mkdocs build --strict` all green. Suite is **690 passed, 1 skipped** in ~48 s (was 687 before the three new tests). The rendered HTML was re-read for the changed docs pages.

PR #603, opened 2026-08-14
Record the v0.1 retirement and v0.2 deployment in the intervention log

JackKelly:
*This PR was written by Claude Code.*

The v0.1 stack was retired from AWS at roughly 19:00 UTC on 13 August 2026 and replaced with v0.2, whose first `live_forecasts` run was the 2026-08-14T00Z slot. This records both, and — after an adversarial review of the first draft — records the NWP outage the v0.1 window actually contained, which the page had been about to describe as a clean sweep.

## The v0.1 window was not clean

The first draft of this PR carried the page's existing claim of "zero interventions and zero observed failures" forward across the closed window. That claim was wrong, and the production Delta tables show it:

- **The 9 August `ecmwf_ens` run produced nothing.** The NWP table's history jumps from version 24 (`2026-08-08T08:31:39Z`) straight to version 25 (`2026-08-10T08:31:30Z`). Every other day in the window has an ~08:31 UTC write.
- **Four forecast slots ran degraded.** 2026-08-09 12:00 through 2026-08-10 06:00 UTC all forecast from the 2026-08-08 00Z run, at 36/42/48/54 hours old. Every one of the other 113 slots used NWP no older than 30 hours.
- **A human re-materialised the partition on 13 August at 19:56 UTC.** Version 29 is the only write in the NWP table's entire history carrying a predicate, and it targets exactly `init_time = 2026-08-09T00:00:00`.

The cause was an upstream partial publication: Dynamical.org publishes each run as ~40 Icechunk commits, and this one was first published with a weather variable wholly missing, then repaired 3h25m later. v0.1 treated that as fatal. #493 has since made it a retryable "not ready yet" with a four-hour budget, which covers the observed republication delay.

**This makes the window better evidence for H1 than a clean sweep would have been.** A missing input degraded the forecast instead of stopping it — Principle 1 working in production rather than on paper — and the degradation was bounded and reconstructable after the fact from `nwp_init_time` on each forecast row.

It also exposes a real detection gap, now stated on the page: the Sentry alarm that surfaced the missed run came from pre-v0.2 code running on a laptop, not from the deployed stack, and the missed-check-in monitor never fired.

## What changed

- **The log has its first row** — `2026-08-13 19:56`, `upstream-outage`, `<5` minutes, Runbook? `yes`.
- **Periods table** — v0.1 closed at `2026-08-14 00:00 UTC` with 1 intervention; v0.2 added from that instant, so the backfill does not fall into a gap between the two periods. Time-series counts corrected from 32 to 28 (v0.1) and 31 (v0.2); 32 is the trial-area size, not what the promoted model forecasts.
- **v0.1 section** — retitled to the closed range, verified end to end, and a new subsection on the 9 August run and what it shows.
- **New v0.2 section** — the deployment, the first slot, and the two things that make the next stretch better evidence: the health check, and the NWP retry.

## Verification

Every figure on the page was read from the production `power_forecasts` table in the delivery bucket (`fold_id = "live"`) and the `NWP` table in the internal bucket:

| Claim | Result |
|---|---|
| v0.1 range 2026-07-15 18:00 → 2026-08-13 18:00 UTC | confirmed |
| 117 slots, no gaps, every consecutive pair exactly 6h apart | confirmed |
| 28 time series in every one of the 117 slots | confirmed |
| v0.2: 31 series, 00Z and 06Z on 14 Aug, both carrying all 31 | confirmed; the 28 are a strict subset of the 31 |
| Four slots at 36–54h NWP age, all others ≤30h | confirmed |
| No NWP write on 9 August; predicated backfill at 13 Aug 19:56 UTC | confirmed |

`pymarkdown scan` and `mkdocs build --strict` pass, and the rendered HTML was read back — per the `mkdocs-authoring` skill, both linters can pass on markdown that renders visibly wrong.

Two figures rest on the maintainer's account rather than on data, and both are hedged in the page: the retirement at "roughly 19:00 UTC" (corroborated by the v0.2.0 version-bump merge at 19:00:29 UTC) and the intervention costing about a minute.

Docs only; no code changed. One adversarial sub-agent reviewed the diff; its findings are triaged in the comments below.

### DIFF
--- a/docs/live_service/intervention-log.md
+++ b/docs/live_service/intervention-log.md
@@ -78,10 +78,9 @@ the plainest possible case of the selective reading these hypotheses exist to pr
 
 ## The log
 
-*No interventions recorded yet.*
-
 | Date | Trigger | Cause | Minutes | Runbook? | Notes |
 |---|---|---|---|---|---|
+| 2026-08-13 19:56 | Sentry alarm — but from the pre-v0.2 code on a laptop, not from AWS | `upstream-outage` | <5 | yes | Dynamical.org first published the 2026-08-09 00Z ECMWF run with a variable wholly missing, and v0.1 treated that as fatal, so the partition was re-materialised by hand. Four forecast slots ran on the previous day's run in the meantime. [#493](https://github.com/openclimatefix/nged-substation-forecast/pull/493) added the retry that covers it |
 
 ## Periods covered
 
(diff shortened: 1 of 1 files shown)


JackKelly:
*Comment written by Claude Code.*

Pushed a second commit that replaces the three reviewer notes above with verified figures, read from the production `power_forecasts` Delta table in the delivery bucket (`fold_id = "live"`, grouped by `experiment_name`).

| | v0.1 (`xgboost_cv_0001`) | v0.2 (`xgboost_cv_0003`) |
|---|---|---|
| Time series | 28 | 31 |
| Slots | 117 | 2 so far (00Z, 06Z on 14 Aug) |
| Range | 2026-07-15 18:00 → 2026-08-13 18:00 UTC | from 2026-08-14 00:00 UTC |
| Gaps | none — every consecutive pair exactly 6h apart | — |
| Series per slot | all 28 in every one of the 117 | all 31 in both |

**The page said 32 time series and that was wrong.** v0.1's promoted model forecasts 28. 32 is the trial-area size, so the two numbers had been running together; worth a glance at whether the four-substation gap is the eligibility filter doing its job or something quietly dropping series, but that is outside this PR.

The v0.1 window is now verified end to end, so the hedge about the last 26 slots is gone and "every expected forecast exists" is back — this time meaning all 117 slots present, no gaps, and full series coverage in each, rather than just "distinct slot count matched".

One thing worth knowing for future checks: the *local* Delta table is not the production record. It holds an `xgboost_cv_0001` run from 4–22 July and an `xgboost_cv_0002` run from 7–13 August that never went to AWS at all, because the local service only runs when new code is being tested. Production lives in the delivery bucket, not `DATA_PATH_INTERNAL` — the internal root has only `NGED/` and `NWP/` under it.

Two figures still rest on Jack's account rather than on data, and both are hedged in the page: the retirement at "roughly 19:00 UTC", and the "as of 08:00 UTC" stamp. The 19:00 figure sits an hour after the last v0.1 slot and before the first v0.2 one, so the data is at least consistent with it.

JackKelly:
*Comment written by Claude Code.*

Triage of the adversarial review. Nine findings, seven accepted, two rejected. I re-verified findings 1 and 2 against the Delta tables myself before acting on them rather than taking the reviewer's word.

**1 & 2 — the missed 9 August NWP run and the hand-backfill. Accepted, and they are the reason this PR grew.** Both confirmed independently: NWP history skips 9 August, four slots at 36/42/48/54h NWP age against ≤30h for the other 113, and version 29 is the only predicated write in the table's history. The page now carries a log row and a subsection instead of "zero observed failures". The reviewer was right that the first draft would have laundered a real degradation into a clean record on the one page whose purpose is to make that impossible.

One correction to the reviewer's framing: it read the four degraded slots as damage. They are the design working — Principle 1, the forecast never stops — and the page now says so. The genuine finding underneath is the *detection* gap, not the degradation.

**3 — "as of 08:00 UTC" was in the future. Accepted.** Now 07:00 UTC.

**4 — the PR body misdescribed the diff. Accepted.** Rewritten.

**5 — "the first caveat no longer applies" overclaimed. Accepted.** `live_forecasts_are_healthy` is `WARN` and non-blocking, so it alerts nobody. The page now says both v0.2 changes narrow the caveat without closing it.

**6 — "every slot checked". Accepted.** Reworded to "reporting on each slot", and the prose says a slot whose run raised fails outright rather than being checked.

**8 — "arrives with v0.2" was the wrong tense. Accepted**, reverted to "landed in".

**7 — repetition between the table and the prose. Rejected.** The v0.2 prose does repeat the table's dates and series count, but the periods table is an index and the sections are the account; a reader arriving at either should not have to hold the other in their head. The paragraphs that carried no information beyond the table are gone regardless, absorbed into the rewrite.

**9 — the table's unhedged 19:00 boundary versus the prose's "roughly". Rejected as stated, fixed by other means.** The v0.1 period now ends at `2026-08-14 00:00 UTC` — when v0.2 took over — so the boundary no longer asserts a retirement instant at all, and the two periods are contiguous with the backfill inside v0.1's rather than in a gap. Separately, the ~19:00 figure is now corroborated: the v0.2.0 version-bump merge landed at 19:00:29 UTC.

**What the reviewer could not check, and still cannot be:** the Dagster run history and the Sentry monitor state, neither reachable from here. Given finding 1 I dropped the page's claim that the Sentry monitor "never alarmed" as a *supporting* cross-check — for AWS it is now stated as the detection gap it actually is.

Issue #501, opened 2026-08-10
Warning checks don't reach Sentry, so degradation lands silently


JackKelly:
*Filed by Claude Code, at Jack's direction, from an adversarial review of [#493](https://github.com/openclimatefix/nged-substation-forecast/pull/493). Widened on 2026-08-14 from "a near-empty NWP variable lands silently" to cover every warning check, after the 9 August NWP outage showed the same silence on a second check.*

## The design rule this serves

The code should always attempt to deliver a power forecast, even when its inputs degrade — but **we must be notified about the degradation**, through three mechanisms:

1. **Wider confidence bands on the forecast itself**, so the degradation is visible to whoever reads the forecast. Designed, not yet built — see [Widening bands](https://openclimatefix.github.io/nged-substation-forecast/design-philosophy/inherent-stability/#widening-bands-the-in-band-signal).
2. **A row in the `power_forecast_warnings` table**, so the degradation is queryable per forecast. Not yet built — [#439](https://github.com/openclimatefix/nged-substation-forecast/issues/439) is Phase 1, and its `STALE NWP` warning type is exactly this case.
3. **A Sentry event, for data failures a human can help with** — for example, no NWP downloaded for over a day, which says Dynamical.org is having problems and somebody should look.

This issue is about the third mechanism. Mechanisms 1 and 2 have their own issues; all three are independent, and none substitutes for another.

## The gap

**No warning check reaches Sentry.** Two checks detect real degradation, warn, and tell nobody outside the Dagster UI.

Both were previously explained away with the same wrong reason — that Sentry fires on a failed *run*, so a non-blocking `WARN` check cannot notify. **That is not true.** `power_data_is_fresh` already does exactly this: on its normal path it calls `report_power_freshness` (`checks.py:399`), which sends a `warning`-level Sentry event whenever the result is unhealthy, fingerprinted per environment so repeated reports collapse into a single Sentry issue. The run still passes and the check is still non-blocking. The pattern exists, works in production, and is simply not applied to the other two checks. That correction landed in [#604](https://github.com/openclimatefix/nged-substation-forecast/pull/604).

### Check 1 — `nwp_has_no_unexpected_nulls` (`n_whole_null_h3_slices`)

[...]

PR #609, opened 2026-08-14
Make notifying us about degradation a required half of Principle 1

JackKelly:
*This PR was written by Claude Code.*

Records the design rule: the code should always attempt to deliver a power forecast even when its inputs degrade, **but we must be notified about the degradation** — through wider confidence bands on the forecast, a row in the `power_forecast_warnings` table, and a Sentry event for data failures a human can act on. Adds the distinction between a Sentry event meaning "we broke" and one meaning "an input degraded", and requires both to be delivered.

## Why this is a change and not a restatement

The first half was already there. The second half was written as optional. Rule 4 in [Inherent Stability](https://openclimatefix.github.io/nged-substation-forecast/design-philosophy/inherent-stability/#the-rules) read:

> **Signal degradation in-band first.** … Side channels — warning tables, checks, Sentry — supplement it; they never substitute for it.

"Supplement" permits a degradation that reaches nobody so long as the band widened. That is what happened on 9 August: the forecast degraded gracefully across four slots and nothing on AWS said a word.

## What changed

- **Principle 1** — degrading is now half the principle, with all three channels named and cross-linked, and notification stated as required.
- **Principle 1's "Decided" line** — *non-blocking never means non-notifying*: a check that detects degradation must still send a Sentry event, without failing the run it is warning about.
- **Rule 4** — now "Signal degradation on all three channels, in-band first". The band keeps priority in ordering, not in obligation.
- **Rule 6** — carries the non-blocking/non-notifying distinction too, since the rules section is meant to be self-contained.
- **New "Two kinds of Sentry event"** — the split, what separates them, and that both must be delivered.
- **The failure-modes table** — the alert column now names which kind of event fires. Three rows were factually wrong about today; the remaining degradation rows no longer read as intended silence.
- **The page's opening summary** and the **"Three audiences"** developer row — both stated only part of the picture.

## The Sentry split

| | Level | Carries | Means | Sent by |
|---|---|---|---|---|
| **Something broke** | `error` | an exception | our code, or something it depends on, could not do its job | `sentry_capture_failure`, `report_asset_degradation`, `report_check_degradation` |
| **Something degraded** | `warning` | a message and its context | nothing threw; an input is late, stale or thin, and the forecast carried on | `report_power_freshness` |

The exception is the dividing line, and it is exact: an error event always carries one, a warning event never does. Crucially **an error event does not mean the run died** — two of its three senders exist precisely because their callers caught the exception and carried on. The missed-check-in alarm is neither kind: it is Sentry noticing a heartbeat that did not arrive.

The two sides are unevenly built. Three senders cover "broke"; "degraded" has one, which is why most of the degradation our checks detect is invisible outside Dagster — #501.

## Review history

Two independent adversarial sub-agent reviews, neither of which came back green; both are triaged in the comments. Between them they caught, and this branch corrects:

- A false causal claim — that the checks not reaching Sentry is *why* a degraded slot sends a healthy check-in. The heartbeat is sent from the asset body before any check runs, and its success-only behaviour is deliberate.
- "Report only to Dagster", which was false: every check reaches Sentry via `report_check_degradation` when the check itself breaks.
- The fatal/non-fatal labelling, which inverted the purpose of two senders that exist *because they never fail a run*.
- Two rows claiming a Sentry error event that cannot fire: the failure hook is on three scheduled jobs and `promoted_model` is in none of them.
- "No NWP downloaded for over a day" as the example, which rule 5 forbids — healthy NWP is already 12–30 hours old, so that threshold fires daily. Now "a missed daily ECMWF run".
- A new heading inserted mid-section, orphaning paragraphs that four pages and a docstring link to.

## Verification

`uv run pymarkdown scan` and `uv run mkdocs build --strict` pass on both files. Both new anchors — `two-kinds-of-sentry-event` and `three-audiences-three-channels` — were grepped out of the built HTML, and Principle 1, the Sentry table and the failure-modes table were each read back from `site/` rather than assumed. Every "Human alerted?" cell was traced to the code: which jobs carry `sentry_capture_failure` (`defs/schedules.py:19,46,70` — three, and no others), whether the asset raises, and which sender fires.

Docs only; no code changed.

### DIFF
--- a/docs/design-philosophy/design-principles.md
+++ b/docs/design-philosophy/design-principles.md
@@ -76,10 +76,12 @@ our forecast will *always* be better than NGED's incumbent baseline, even when w
 data.)
 
 Degrading is only half the principle: **we must be notified that the forecast degraded.** Three
-channels carry that — the widened uncertainty band on the forecast row, a `power_forecast_warnings`
-row naming which feed degraded and since when, and a Sentry event for the data failures a human can
-act on, such as a missed daily ECMWF run that says Dynamical.org is having problems. Each answers a
-different question and none substitutes for another; all three are required. What each is for, and
+channels carry that — a widened uncertainty band on the forecast row (designed, not yet built 🚧 —
+see [Widening bands](inherent-stability.md#widening-bands-the-in-band-signal)), a
+`power_forecast_warnings` row naming which feed degraded and since when (not yet built 🚧), and a
+Sentry event for the data failures a human can act on, such as a missed daily ECMWF run that says
+Dynamical.org is having problems. Each answers a different question and none substitutes for
+another; all three are required, and two of the three are still to be built. What each is for, and
 who reads it, is set out in
 [Three audiences, three channels](inherent-stability.md#three-audiences-three-channels). A Sentry
 event says whether we broke or an input degraded, and both kinds have to be delivered — see
--- a/docs/design-philosophy/design-principles.md
+++ b/docs/design-philosophy/design-principles.md
@@ -65,13 +65,27 @@ principle behind it is a claim we are merely hoping comes true.
 
 ### 1 — The power forecast never stops
 
-If data inputs are disrupted, the forecast gets less certain instead of stopping — and should say
-so in the forecast itself, through wider uncertainty bands (band-widening is designed but not yet
-built — see [Widening bands](inherent-stability.md#widening-bands-the-in-band-signal)). The
+If data inputs are disrupted, the forecast gets less certain instead of stopping. The
 forecast always does the best it can with whatever data it has, rather than blowing up; raising
 is reserved for states that are our own bug. The plan is to deliver that through the **model
 itself** — an ML model that can, at least partially, handle missing inputs — rather than through
-fallback logic wrapped around a model that assumes complete data. (Note that this decision to
+fallback logic wrapped around a model that assumes complete data.
+
+Degrading is only half the principle: **we must be notified that the forecast degraded.** Three
+channels carry that, each answering a different question, and none substitutes for another —
+see [Three audiences, three channels](inherent-stability.md#three-audiences-three-channels).
+
+- **Wider uncertainty bands on the forecast itself**, so whoever reads a row can tell how much to
+  trust it. Designed, not yet built — see
+  [Widening bands](inherent-stability.md#widening-bands-the-in-band-signal).
+- **A row in `power_forecast_warnings`**, naming which feed degraded and since when, so a data
+  provider's problem is attributable to that provider. Not yet built — see
+  [Delivery tables](../roadmap/delivery-tables.md#table-2-power_forecast_warnings).
+- **A Sentry event, for data failures a human can help with** — no NWP downloaded for over a day,
+  say, which says Dynamical.org is having problems and somebody should look. Partly built:
+  `power_data_is_fresh` sends one, the other warning checks do not.
+
+(Note that this decision to
 "never stop" will not be appropriate for energy-forecasting systems where an uncertain forecast
 might be more harmful than *no* forecast. But, in Flexpectation, there are strong arguments that
 our forecast will *always* be better than NGED's incumbent baseline, even when we have no live
@@ -82,7 +96,8 @@ because one meter went quiet, NGED open their dashboard to a gap instead of a fo
 developer spends the morning re-running a pipeline whose only real problem was a missing input.
 
 *Decided:* every asset check in the repo is non-blocking `WARN`; there is deliberately no
-`ERROR`-severity check anywhere.
+`ERROR`-severity check anywhere. Non-blocking never means non-notifying — a check that detects
+degradation still has to raise a Sentry event, it just must not fail the run.
 
 *Serves:* [Hypothesis 1: a service that mostly runs
 itself](engineering-hypotheses.md#h1-a-service-that-mostly-runs-itself).
--- a/docs/design-philosophy/design-principles.md
+++ b/docs/design-philosophy/design-principles.md
@@ -69,7 +69,11 @@ If data inputs are disrupted, the forecast gets less certain instead of stopping
 always does the best it can with whatever data it has, rather than blowing up; raising is reserved
 for states that are our own bug. The plan is to deliver that through the **model itself** — an ML
 model that can, at least partially, handle missing inputs — rather than through fallback logic
-wrapped around a model that assumes complete data.
+wrapped around a model that assumes complete data. (Note that this decision to
+"never stop" will not be appropriate for energy-forecasting systems where an uncertain forecast
+might be more harmful than *no* forecast. But, in Flexpectation, there are strong arguments that
+our forecast will *always* be better than NGED's incumbent baseline, even when we have no live
+data.)
 
 Degrading is only half the principle: **we must be notified that the forecast degraded.** Three
 channels carry that — the widened uncertainty band on the forecast row, a `power_forecast_warnings`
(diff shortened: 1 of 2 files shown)


JackKelly:
*Comment written by Claude Code.*

An adversarial sub-agent reviewed this branch and returned twelve confirmed findings plus two plausible. Ten accepted, two rejected. I re-verified the two that change meaning against the source before acting. `main` is also merged in.

**The worst one — a false causal claim I introduced.** The page said the checks "report only to Dagster, **so** a degraded slot still sends a healthy check-in". The "so" is wrong. `send_forecast_checkin` is called from the `live_forecasts` asset body after a successful write (`production_assets.py:326`), gated only on `availability_mode == "live"` — it runs *before* any check does, so wiring Sentry into the checks would not change the check-in by one bit. Worse, the success-only heartbeat is deliberate: the alarm is meant to fire purely on the *absence* of a heartbeat, so the sentence implied a fix that would break the alarm. Now stated as two independent facts, with the second marked deliberate.

**"Report only to Dagster" was also false.** Every check does reach Sentry through `report_check_degradation` — `checks.py:1016` and `assets.py:446`, the latter shared by all three `ecmwf_ens` checks. The gap is narrower and stranger than the page claimed: they notify when they *break*, and never when they work and find the degradation they exist to find. Confirmed by reading the guard at `assets.py:440-450`.

**Rule 5 caught the example.** "No NWP downloaded for over a day" is an absolute age threshold, and rule 5 four lines away forbids measuring degradation in hours of age — healthy NWP is already 12 to 30 hours old depending on the slot, so a one-day threshold fires on the 06:00 slot every day. Reworded to "a missed daily ECMWF run", which is the same fault, is what `n_missed_nwp_runs` actually measures, and keeps the intent.

**A pre-existing error the change exposed.** The failure-modes table said no human is alerted when telemetry stalls for one series — but `power_data_is_fresh` has always sent a Sentry warning, and `sentry.md` tells you to add an alert rule for it. Corrected to "Yes, next business day". That cell was wrong before this branch; the new text just made it contradict a neighbour.

Also accepted: the provider channel is not wholly unbuilt (only `power_forecast_warnings` is — the late-series table, missed-run count and null-rate metadata all exist); "non-blocking never means non-notifying" now appears in rule 6 as well, since the rules section is meant to be self-contained; the page's opening summary named only the in-band half; the widening-bands claim that "no separate monitoring system has to work" is now scoped to the consumer-facing property, so it stops reading as an argument that the other two channels are optional; two line-wrap artifacts; a comma splice; and "raise" swapped for "send", since "raise" is load-bearing on these pages and the principle is about *not* raising.

**Rejected — rule 4 carrying the same principle marker as rule 1.** The reviewer argued two rules should not both claim to be the imperative form of Principle 1. The section's own convention says rules that restate a principle are marked so they can be changed together, and Principle 1 is described on that page as "the first and largest". Rule 1 is the degrade half, rule 4 the notify half; both genuinely derive from it and both should move if it does.

**Rejected — that the intervention log now overclaims** in calling the 9 August episode "Principle 1 working in production". The very next caveat on that page says nothing alerted us from AWS, so the qualification a reader needs is already adjacent.

**One judgement call left for the maintainer, deliberately not made here.** Six rows of the failure-modes table read "Intended: Unchanged" with "Human alerted? No" — for missed NWP runs, stale NWP, an empty slot, and so on. Under the new rule those intentions are wrong: notification is now required, so the intended state for each is an alert. Rewriting six rows of a table nobody asked me to touch is a scope decision rather than a correction, so the table is left as it is and this is the flag.

JackKelly:
*Comment written by Claude Code.*

Second adversarial review, by a fresh sub-agent. Eight findings, all eight accepted. I re-verified the two structural ones against source before acting. The PR body has also been rewritten — it was still asserting the causal claim retracted two commits ago.

**The framing was wrong at its centre.** I had labelled the split "Fatal" versus "Worrying, not fatal", and put `report_asset_degradation` and `report_check_degradation` under "Fatal". `_sentry.py:12-14` says they "cover the production faults the hook cannot see, **because they never fail a run**: a check, or an asset, that caught its own exception instead of raising." So they are the caught-and-carried-on case, and filing them under "Fatal" inverted their purpose. I had also claimed the split was "the same distinction rule 1 draws between raising and degrading" — it is not. It separates whether an exception was available to attach, which cuts *across* that line: the metadata-roster row degrades gracefully and still sends an error event.

Relabelled to **"Something broke"** and **"Something degraded"**, with an explicit paragraph that an error event does not mean the run died.

**Two rows claimed an alert that cannot fire.** I had written "Yes — **error**, at promotion time" on both promotion rows. `sentry_capture_failure` is attached to exactly three jobs (`defs/schedules.py:19,46,70`; `grep -rn "hooks=" src/ packages/` returns nothing else), and `promoted_model` is in none of them — it is a bare `@asset` materialised by hand from the Dagster UI, and `sentry.md:138-141` already says ad-hoc materialisations never reach Sentry. The original cell, "Yes, at promotion time", was true because the operator is watching the launchpad; adding the word "error" made it false. Now says both halves.

**A heading in the wrong place.** Inserting `### Two kinds of Sentry event` mid-section orphaned the `warning_source` justification and the whole 14-hour-deadline argument under a heading about event levels — content that `delivery-tables.md`, `production-deployment.md`, `inherent-stability.md` itself and a docstring in `checks.py:36` all link to as part of "Three audiences, three channels". The anchors still resolved, which is why the build check missed it. Moved after that section closes.

Also accepted: the missed-check-in alarm is neither kind of event, so a legend asserting every row resolves to one of two was wrong — it is now named as the third thing it is; the audience table's developer row asked "Is *our* system at fault?", which contradicts the new text saying Sentry also carries upstream problems; the roster row alerts only when the upsert *raises*, not when the roster merely loses rows, which raises nothing; the meter row pointed at #501, which would not change it; and the never-stop caveat had been left as a free-floating parenthetical two paragraphs from its referent.

**Nothing was rejected this round.** Every finding held up.

**What the reviewer verified and found sound**, so a reader knows what the corrections sit on: exactly four explicit senders, correctly named on each side; the claim that an error event always carries an exception and a warning event never does — checked empirically against the locked sentry-sdk rather than by reading, including that log-to-event capture is disabled so the `logger.exception` calls in the guards cannot emit events; the ECMWF row, including that the failure hook unwraps `RetryRequested` to its `__cause__`; the check-in ordering; and that no other page in the repo still carries the old framing.

JackKelly:
*Comment written by Claude Code.*

Third adversarial review, fresh sub-agent. Five confirmed findings and three plausible; all eight accepted. **Four of the five confirmed were introduced by the previous two correction rounds, not by the original change** — that pattern is the most useful thing this round produced, and it is noted at the bottom.

**A status marker lost to an earlier fix.** Review 1 objected that the principles page is "an index, not the argument", so I collapsed the three-channel bullet list into prose — and the build-status markers went with it. `design-principles.md` was left asserting in the present tense that three channels carry the notification, with zero 🚧 anywhere on the page and no remaining link to the section saying two of the three do not exist. Verified: `grep -c "🚧\|not yet built" docs/design-philosophy/design-principles.md` returned 0. Restored inline.

**"The run stays green and the failure hook never fires" was false.** Both senders are called part-way through an asset with a Delta write still ahead of it — `assets.py:182` before the power write, `assets.py:446` before `write_nwp`. If that write raises, a degradation event and a run-failure event land on the same run. The preceding clause was true and sufficient; the absolute I added on top was not. Now says so, and points at `fault_category:run_failed`, which is what `operations.md` already triages on.

**Two rows contradicted a row three below them.** "One or two daily NWP runs missed" answered "Not yet"; "A whole ECMWF weather variable absent" says the exhausted retry ladder fails the run and fires an error event, then adds that it "manifests downstream as a missed run". Same chain, opposite answers. The second is right — `ecmwf_ens_job` carries the hook and `sentry_capture_failure` unwraps `RetryRequested` to its `__cause__`. #501's real scope is the narrower claim that the *check's* missed-run count does not reach Sentry. Both rows now carry both halves.

**A row that contradicted itself.** The meter row said "Partly detected at ingest" in one cell and "the detection itself is unbuilt" in the other. `PowerTimeSeries.drop_implausible_rows` drops malformed timestamps and detects no wrong *values*, so the row now says that in both cells.

**A lead-in written for a table that had since changed.** It enumerated three possible answers; the promotion rows, corrected last round to "the operator is watching the launchpad. No Sentry event", use a fourth. Admitted.

Also accepted: rule 6 now carries the principle marker its own section's convention requires; and the opening blockquote, which had gained a second unbuilt channel stated as fact, is hedged.

**Nothing rejected.**

**Worth recording:** three rounds, three red verdicts, and the defect count did not converge — round 3 found five confirmed, four of them regressions from rounds 1 and 2. Each round's fixes were individually sound and collectively introduced new inconsistencies, because this change touches a table, a taxonomy and two summary paragraphs that all have to agree. That is a signal about the shape of the change, not about the reviewers.

